PR #1, opened 2023-08-09
Add files via upload

MartinDaCosta53:


MartinDaCosta53:
These are the files that I used successfully this morning.  Since then, I've broken my installation.  Can't work out what I've done yet.

MartinDaCosta53:
Yes, this will change but one step at a time!

src/Controller.cpp
```diff
@@ -81,101 +92,69 @@ void Controller::setName(unsigned char *mname)
...
-  setSLiM();
+  return (amsg->rtr);
 }
 
 //
-/// check 30 sec timeout for SLiM/FLiM negotiation with FCU
+/// if in FLiM mode, initiate a CAN ID enumeration cycle
 //
-void Controller::checkModeChangeTimeout()
+void Controller::startCANenumeration()
```
SvenRosvall, line 139: Hmm, seems like you are out-of sync here as well.

src/MinimumNodeService.cpp
```diff
@@ -57,11 +57,120 @@ void MinimumNodeService::setController(Controller *cntrl)
 // * SQU
 // * NNRST
 // * NNRSM
+
```
SvenRosvall, line 60: The comments above here relate to the method handleMessage(). Please move these comments down to just before this method.

src/MinimumNodeService.cpp
```diff
@@ -149,6 +258,46 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
 
       return PROCESSED;
 
+    case OPC_CANID:
```
SvenRosvall, line 261: This is again something I did earlier. I have since moved it to the CanService.
Again, I believe an update from main branch will fix this.

PR #3, opened 2023-08-12
Initialise Factory Reset EEPROM ready for use

MartinDaCosta53:


src/Configuration.cpp
```diff
@@ -426,7 +435,7 @@ void Configuration::cleareventEEPROM(byte index)
 #endif
 
 #ifndef AVR_RESET_METHOD
-#define AVR_RESET_METHOD 3     // don't use watchdog timer method as it's unreliable on some boards
+#define AVR_RESET_METHOD 4     // don't use watchdog timer method as it's unreliable on some boards
```
SvenRosvall, line 438: Is this reset method better? In what way?

MartinDaCosta53: When I first started to look at the UI system with Duncan, he was using Reset Method 1.  I found this caused the system to hang.  Going down the list, number 3 was the first that worked, which is why it is selected in Duncan's code as the default.  Whilst looking at EEPROM initialisation, I looked to call the resetModule() function, which resets the processor at the end.  The system hung on re-start.  Rest method 4 works. Pointing to this interrupt vector seems to be the method used if you go on the web for Arduino reset advice.

Subsequently, I decided not to call resetModule() as it made too many unnecessary EEPROM writes and I saw no reason for reset here.

PR #6, opened 2023-08-16
Heartbeat and multiple setup prevention

MartinDaCosta53:


src/MinimumNodeService.cpp
```diff
@@ -96,6 +96,20 @@ void MinimumNodeService::checkModeChangeTimeout()
   }
 }
 
+void MinimumNodeService::heartbeat()
+{
+  if ((module_config->currentMode == MODE_NORMAL) && !noHeartbeat)
+  {
+    if ((millis() - lastHeartbeat) > heartRate)
```
SvenRosvall, line 103: These two if-s can be combined. But actually I think it is easier to read this way.
I don't think combining them will yield different machine code.

MartinDaCosta53:
Yes, I thought about it but left as is for the reason you stated.

PR #7, opened 2023-08-17
Simplify factory/virgin memory conditioning to include setting NVs to…

MartinDaCosta53:
… zero

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -49,18 +49,15 @@ void Configuration::begin()
   EE_BYTES_PER_EVENT = EE_NUM_EVS + 4;
 
   storage->begin();
-
-  makeEvHashTable();
-  loadNVs();
   
   if ((currentMode == 0xFF) && (nodeNum == 0xFFFF))   // EEPROM is in factory virgin state
   {
-   storage->writeEEPROM(0, 0);     // SLiM
-   storage->writeEEPROM(1, 0);     // CANID
-   storage->writeEEPROM(2, 0);     // NN hi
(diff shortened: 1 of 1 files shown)


dpharris:
Should "makeEvHashTable();" exist here, or only in ConsumerEvent code?

MartinDaCosta53:
Probably but we haven’t written that service yet.  There may well be code that migrates from here as well as CbusService when we do

SvenRosvall:
Thanks David for reviewing and commenting here. Your comments help us (and hopefully also you) understand the code better.

It is tricky to decide where things go, which class they belong to. makeEvHashTable() is such an example. The Configuration class deals with anything that needs to be persisted. It knows what NVs and EVs are and how to translate each such piece of data into an EEPROM address. The hash table is part of the storage code to make EEPROM access more efficient. I don't think it will belong in the event services for this reason. I agree that this code borders between pure events and the storage of them. 
There might be a different way to map event data to EEPROM storage that we come up with some time later.

dpharris:
Thanks.  Because the Storage holds both NVs and Events, you have defined a
common method of assigning their storage locations.

To separate into NV-Service and Event-Service one would need some way to
communicate the range of Storage available for each Service.

What I did in my Toy-code was to keep a global storage_location variable
which was passed to each Service.  Each Service would 'reserve' its
required storage, and pass back the updated storage_location.  This allows
the NV space to expand as the developer wants, the event-space moves up.
Each Service retains the base of its range, and uses that to calculate the
storage-address of the item.

David

PR #8, opened 2023-08-18
Add OPC_NNRSM & OPC_NNRST

MartinDaCosta53:
Adds support for:
Reset to manufacturer's default, &
Software reset


dpharris:
Nice.  I like how you have set things up.

I looked at configuration::reset() -- ugh for the variations needed ...
better you than me figuring that out!

David

PR #9, opened 2023-08-18
Create EventConsumerService

MartinDaCosta53:
Op-codes and functions moved from CbusService.  Examples updated.

### DIFF
--- a/src/EventConsumerService.cpp
+++ b/src/EventConsumerService.cpp
@@ -1,0 +1,122 @@
+// Copyright (C) Sven Rosvall (sven@rosvall.ie)
+// This file is part of VLCB-Arduino project on https://github.com/SvenRosvall/VLCB-Arduino
+// Licensed under the Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License.
+// The full licence can be found at: http://creativecommons.org/licenses/by-nc-sa/4.0/
+
+#include <Streaming.h>
+
+#include "EventConsumerService.h"
+#include <Controller.h>
+#include <vlcbdefs.hpp>
+
+namespace VLCB {
+
+void EventConsumerService::setController(Controller *cntrl) 
+{
+  this->controller = cntrl;
+  this->module_config = cntrl->module_config;
+}
+
+//
+/// register the user handler for learned events
+//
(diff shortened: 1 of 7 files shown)


src/EventConsumerService.cpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    case OPC_ACON:
+    case OPC_ACON1:
+    case OPC_ACON2:
+    case OPC_ACON3:
+
+    case OPC_ACOF:
+    case OPC_ACOF1:
+    case OPC_ACOF2:
+    case OPC_ACOF3:
```
SvenRosvall, line 70: There is a discussion if these messages with data should be used with the event handler that doesn't handle that data. I am happy to leave as is for now and fix the ones with data when we know what to do with them.

src/EventConsumerService.cpp
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      return PROCESSED;
+
+    case OPC_AREQ:
+      // AREQ message - request for node state, only producer nodes
+
+      if ((msg->data[1] == highByte(module_config->nodeNum)) && (msg->data[2] == lowByte(module_config->nodeNum))) 
+      {
+        (void)(*eventhandler)(0, msg);
```
SvenRosvall, line 105: Just a note that this is the wrong behaviour. AREQ is not an incoming event. It is a request for producer to show its status. No need to fix now. 
I think it should go in the EventProducerService. Maybe put AREQ back into CbusService so we don't forget it when we create the producer service.

src/EventConsumerService.cpp
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      return PROCESSED;
+
+    case OPC_ASRQ:
```
SvenRosvall, line 110: Oops. No implementation. Something for you to do next. :)
But again, this should be in the producer service just like AREQ.

MartinDaCosta53: I've only got in the op-codes that Duncan implemented at the moment.  He didn't implement ASRQ.
I saw Duncan's note about AREQ but the Service specs show both AREQ and ASRQ in Consumer and Producer.

PR #11, opened 2023-08-21
Create Event Teaching Service

MartinDaCosta53:


SvenRosvall:
Please also include the changes for CbusService.cpp . I expect code to be removed here.

MartinDaCosta53:
Sorry, I missed it.  Hopefully it's there now

src/EventTeachingService.cpp
```diff
@@ -0,0 +1,261 @@
...
+  unsigned int nn = (msg->data[1] << 8) + msg->data[2];
+  unsigned int en = (msg->data[3] << 8) + msg->data[4];
+  // DEBUG_SERIAL << ">CbusSvc handling message op=" << _HEX(opc) << " nn=" << nn << " en" << en << endl;
+
+  switch (opc)
+  {
+    
+  case OPC_NNLRN:
+    // received NNLRN -- place into learn mode
+     DEBUG_SERIAL << F("> NNLRN for node = ") << nn << F(", learn mode on") << endl;
```
SvenRosvall, line 31: Please comment out this debug output. Want to keep them out to reduce memory usage.

src/EventTeachingService.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+    if (nn == module_config->nodeNum)
+    {
+      controller->bLearn = true;
```
SvenRosvall, line 35: Sorry, can't add comment for this in Controller.h. Now that you have moved the code here, all uses of bLearn are in this service. Can you move this flag from Controller to the EventTeachingService?
I also see that bModeChanging and timeOutTimer in the controller are not used anymore since you added the MinimumNodeService. Can you remove them from the Controller?
(These are the kind of help I get from CLion, the IDE I talked about. ;) )

MartinDaCosta53:
All done.

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
MartinDaCosta53: Done

PR #13, opened 2023-08-21
Minor cleanups.

SvenRosvall:
A few small things to clean up in code. Comments are included in the PR.

### DIFF
--- a/examples/VLCB_1in1out/VLCB_1in1out.ino
+++ b/examples/VLCB_1in1out/VLCB_1in1out.ino
@@ -82,10 +82,11 @@ VLCB::Configuration modconfig;               // configuration object
 VLCB::CAN2515 can2515;                  // CAN transport object
 VLCB::MinimumNodeService mnService;
 VLCB::CanService canService;
-VLCB::CbusService cbusService;               // service for CBUS op-codes
 VLCB::EventConsumerService ecService;
 VLCB::EventTeachingService etService;
-VLCB::Controller controller(&userInterface, &modconfig, &can2515, { &mnService, &canService, &cbusService, &ecService, &etService }); // Controller object
+VLCB::CbusService cbusService;               // service for CBUS op-codes
+VLCB::Controller controller(&userInterface, &modconfig, &can2515, 
+                            { &mnService, &canService, &ecService, &etService, &cbusService }); // Controller object
 
 // module objects
 VLCB::Switch moduleSwitch(5);            // an example switch as input
--- a/src/CanService.h
+++ b/src/CanService.h
@@ -24,7 +24,7 @@ public:
 
   virtual void process(UserInterface::RequestedAction requestedAction) override;
   virtual Processed handleRawMessage(CANFrame *msg) override;
-  Processed handleMessage(unsigned int opc, CANFrame *msg) override;
+  virtual Processed handleMessage(unsigned int opc, CANFrame *msg) override;
 
   void startCANenumeration();
 
--- a/src/Controller.cpp
+++ b/src/Controller.cpp
@@ -141,15 +141,15 @@ void Controller::process(byte num_messages)
     // at least one CAN frame is available in the reception buffer
     // retrieve the next one
 
-    _msg = transport->getNextMessage();
+    CANFrame msg = transport->getNextMessage();
     // DEBUG_SERIAL << "> Received a message" << endl;
 
-    callFrameHandler(&_msg);
+    callFrameHandler(&msg);
 
     for (Service * service : services)
     {
       // DEBUG_SERIAL << "> Passing raw message to " << endl;
-      if (service->handleRawMessage(&_msg) == PROCESSED)
+      if (service->handleRawMessage(&msg) == PROCESSED)
       {
         // DEBUG_SERIAL << "> Service handled raw message" << endl;
         break;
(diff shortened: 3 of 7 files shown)


examples/VLCB_1in1out/VLCB_1in1out.ino
```diff
@@ -82,10 +82,11 @@ VLCB::Configuration modconfig;               // configuration object
 VLCB::CAN2515 can2515;                  // CAN transport object
 VLCB::MinimumNodeService mnService;
 VLCB::CanService canService;
-VLCB::CbusService cbusService;               // service for CBUS op-codes
 VLCB::EventConsumerService ecService;
 VLCB::EventTeachingService etService;
-VLCB::Controller controller(&userInterface, &modconfig, &can2515, { &mnService, &canService, &cbusService, &ecService, &etService }); // Controller object
+VLCB::CbusService cbusService;               // service for CBUS op-codes
+VLCB::Controller controller(&userInterface, &modconfig, &can2515, 
+                            { &mnService, &canService, &ecService, &etService, &cbusService }); // Controller object
```
SvenRosvall, line 89: Wrap long line. Also put cbusService last as it is a catch-all service.

src/CanService.h
```diff
@@ -24,7 +24,7 @@ class CanService : public Service
 
   virtual void process(UserInterface::RequestedAction requestedAction) override;
   virtual Processed handleRawMessage(CANFrame *msg) override;
-  Processed handleMessage(unsigned int opc, CANFrame *msg) override;
+  virtual Processed handleMessage(unsigned int opc, CANFrame *msg) override;
```
SvenRosvall, line 27: Marking all overridden methods with "virtual" so it is easier to find the virtual methods in a class.

src/Controller.cpp
```diff
@@ -141,15 +141,15 @@ void Controller::process(byte num_messages)
     // at least one CAN frame is available in the reception buffer
     // retrieve the next one
 
-    _msg = transport->getNextMessage();
+    CANFrame msg = transport->getNextMessage();
```
SvenRosvall, line 144: Making the message a local variable instead of a class field. It is only used here.

src/Controller.h
```diff
@@ -69,17 +69,16 @@ class Controller
   void setFrameHandler(void (*fptr)(CANFrame *msg), byte *opcodes = NULL, byte num_opcodes = 0);
 
 private:                                          // protected members become private in derived classes
-  CANFrame _msg;
```
SvenRosvall, line 72: See above. Changed this to a local variable.

src/Controller.h
```diff
@@ -69,17 +69,16 @@ class Controller
...
   Transport * transport;
 
   unsigned char *_mparams;
   unsigned char *_mname;
-  void (*framehandler)(CANFrame *msg);
-  byte *_opcodes;
-  byte _num_opcodes;
+  void (*framehandler)(CANFrame *msg) = NULL;
+  byte *_opcodes = NULL;
+  byte _num_opcodes = 0;
```
SvenRosvall, line 81: While working on test cases I noticed a few cases where these fields were not null/zero which caused problems. 
All other fields are set when the controller is created.

src/LongMessageService.h
```diff
@@ -41,9 +41,9 @@ class LongMessageService : public Service
 public:
 
   virtual void setController(Controller *cntrl) override { this->controller = cntrl; }
+  virtual Processed handleMessage(unsigned int opc, CANFrame *msg) override;
```
SvenRosvall, line 44: Mark "virtual". See above. And move up so that all virtual methods are kept together.

src/LongMessageService.h
```diff
@@ -59,9 +59,11 @@ class LongMessageService : public Service
...
   bool _is_receiving = false;
   byte *_send_buffer, *_receive_buffer;
-  byte _send_stream_id = 0, _receive_stream_id = 0, *_stream_ids = NULL, _num_stream_ids = 0, _send_priority = DEFAULT_PRIORITY, _msg_delay = LONG_MESSAGE_DEFAULT_DELAY, _sender_canid = 0;
-  unsigned int _send_buffer_len = 0, _incoming_message_length = 0, _receive_buffer_len = 0, _receive_buffer_index = 0, _send_buffer_index = 0, _incoming_message_crc = 0, \
-                                  _incoming_bytes_received = 0, _receive_timeout = LONG_MESSAGE_RECEIVE_TIMEOUT, _send_sequence_num = 0, _expected_next_receive_sequence_num = 0;
+  byte _send_stream_id = 0, _receive_stream_id = 0, *_stream_ids = NULL, _num_stream_ids = 0;
+  byte _send_priority = DEFAULT_PRIORITY, _msg_delay = LONG_MESSAGE_DEFAULT_DELAY, _sender_canid = 0;
+  unsigned int _send_buffer_len = 0, _incoming_message_length = 0, _receive_buffer_len = 0, _receive_buffer_index = 0;
+  unsigned int _send_buffer_index = 0, _incoming_message_crc = 0, _incoming_bytes_received = 0;
+  unsigned int _receive_timeout = LONG_MESSAGE_RECEIVE_TIMEOUT, _send_sequence_num = 0, _expected_next_receive_sequence_num = 0;
```
SvenRosvall, line 66: Split exceedingly long lists of field definitions.

examples/VLCB_1in1out/VLCB_1in1out.ino
MartinDaCosta53: I was wondering about this earlier.  This line keeps growing as we add more services.  Can we put each service within the braces on its own line? This would make it easier for a user to delete or comment out those they don't need.

SvenRosvall: I was thinking about that too. But I left it like this to avoid too many lines. I thought it would be easy enough to  view the list of services and pick and choose from it as it is shown above.

MartinDaCosta53 (approved):
Looks good to me

PR #16, opened 2023-08-27
Minor fixes.

SvenRosvall:


Arduino/ACAN2515.h
```diff
@@ -42,11 +42,13 @@ struct ACAN2515
   void receive(CANMessage & message);
   bool tryToSend(const CANMessage & message);
   void end();
+  uint16_t receiveBufferCount();
```
SvenRosvall, line 45: This file is supporting unit tests.

src/CanService.cpp
```diff
@@ -11,12 +11,6 @@
 namespace VLCB
 {
 
-enum CanOpCodes
```
SvenRosvall, line 14: I added these OpCodes at an early stage when I thought it would be good to group the OP-codes that belonged to each service. There wasn't much benefit to this idea so cleaning this up now.

src/EventTeachingService.cpp
```diff
@@ -68,7 +68,6 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
 
       if (index < module_config->EE_MAX_EVENTS)
       {
-        // TODO: Review this. j is always 0.
```
SvenRosvall, line 71: This was a TODO I added in Duncan's code. This has then been fixed an "j" has been replaced with "index" and properly set.

src/MinimumNodeService.cpp
```diff
@@ -147,13 +153,6 @@ void MinimumNodeService::process(UserInterface::RequestedAction requestedAction)
      }
    }
 
-// Renegotiating timed out.  Revert to previous NN   
-   if (renegotiating && !bModeSetup)
-   {
-     controller->sendMessageWithNN(OPC_NNACK);
-     setNormal();
-   }
-
```
SvenRosvall, line 156: I want to move this code into checkModeChangeTimeout() as the comment suggests this is to be run at timeout.

As I understand the condition and comment, if bModeSetup has been cleared, we are no longer in setup. This flag is cleared in checkModeChangeTimeout(). So I think we can move this check for renegotiating into checkModeChangeTimeout().

src/MinimumNodeService.cpp
```diff
@@ -416,7 +413,6 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
             break;
           }
           return PROCESSED;
-          break;
```
SvenRosvall, line 419: Removing these breaks as we talked about yesterday.

Arduino/ACAN2515.h
MartinDaCosta53: This file doesn't look anything like the ACAN2515.h that I use.

SvenRosvall: No, this is just a dummy to make unit tests compile. I have only added the necessary declarations here. It does not come from the actual ACAN2515 library (which does not compile outside the Arduino environment).

src/MinimumNodeService.cpp
MartinDaCosta53: Yes, I think that you're correct but we also need 
renegotiating = false; before
controller->sendMessageWithNN(OPC_NNACK); because you're no longer calling setNormal();

Arduino/ACAN2515.h
MartinDaCosta53: OK

src/MinimumNodeService.cpp
SvenRosvall: Thanks. Missed that. Added now.

PR #17, opened 2023-08-28
Check for short messages in accordance with OpCodes Spec para 2.1.1

MartinDaCosta53:


src/MinimumNodeService.cpp
```diff
@@ -212,18 +212,25 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
 
       if (nn == module_config->nodeNum)
       {
-        byte paran = msg->data[3];
+        if (msg->len < 4)
+        {
+          controller->sendGRSP(OPC_RQNPN, getServiceID(), CMDERR_INV_CMD);
```
SvenRosvall, line 217: Should this condition also send a CMDERR?

SvenRosvall: Sorry, I see now that this is a VLCB specific check. So no CMDERR.

MartinDaCosta53: It doesn't seem to be specified in CBUS and Duncan didn't in his code, so I think not.

MartinDaCosta53: Ian doesn't in his code, either

PR #18, opened 2023-08-29
Create NodeVariablesService

SvenRosvall:
Created the NodeVariablesService with code from CbusService (which is now deleted). Error checking implemented as per VLCB specs.


CMakeLists.txt
```diff
@@ -56,6 +54,7 @@ add_library(core_library OBJECT
         src/EventConsumerService.cpp
         src/EventTeachingService.cpp
         src/vlcbdefs.hpp
+        src/NodeVariablesService.cpp
 )
```
SvenRosvall, line 58: This file is for supporting CLion builds.

MartinDaCosta53:
Excellent!  Now we can finally lose the CBUS Service

PR #19, opened 2023-08-29
Pushbutton initiated testing

SvenRosvall:
Added unit tests for changing mode when pressing the push button.
This then resulted in a review of the mode changing code which made me rename some functions.


src/Controller.cpp
```diff
@@ -164,6 +164,8 @@ void Controller::process(byte num_messages)
     _ui->run();
 
     requestedAction = _ui->checkRequestedAction();
+    // if (requestedAction != UserInterface::NONE)
+    //   DEBUG_SERIAL << "Controller::process() UI action=" << requestedAction << " mode=" << module_config->currentMode << endl;
```
SvenRosvall, line 168: This debug output was useful during testing. Leaving it here commented out for future uses.

src/MinimumNodeService.cpp
```diff
@@ -76,20 +76,9 @@ void MinimumNodeService::setUninitialised()
 //
 /// revert from Normal to Uninitialised mode
 //
-void MinimumNodeService::revertUninitialised()
+void MinimumNodeService::initSetupFromNormal()
 {
```
SvenRosvall, line 80: Renamed as we cannot change to Uninitialized anymore.

src/MinimumNodeService.cpp
```diff
@@ -76,20 +76,9 @@ void MinimumNodeService::setUninitialised()
...
-  // send NNREL message
-  controller->sendMessageWithNN(OPC_NNREL);
-  setUninitialised();
-}
-
-//
-/// change or re-confirm node number
-//
-
-void MinimumNodeService::renegotiate()
```
SvenRosvall, line 91: This function is the same as initSetupFromNormal().
Also the term "renegotiate" is misleading as we removed the pushbutton initiated renegotiation functionality.

src/MinimumNodeService.h
```diff
@@ -25,6 +25,7 @@ class MinimumNodeService : public Service
   virtual byte getServiceVersionID() override { return 1; }
   
   virtual void begin() override;
+  void setHeartBeat(bool f) { noHeartbeat = !f; }
```
SvenRosvall, line 28: I needed a quick way of omitting the heart beats while unit testing. The heart beats got in the way.

src/MinimumNodeService.cpp
MartinDaCosta53: I was thinking of removing this function completely as it is not called by anything!  The only reason that I haven't yet is that I was not clear over if we were completely settled on not being able to change to Uninitialised.  My feeling is that we shouldn't be able to make that change other than by a factory reset.

src/MinimumNodeService.cpp
MartinDaCosta53: This function could become initSetupFromNormal().  I kept the original name as referring to renegotiating the node number rather than the original concept of CAN ID.  However, it isn't really renegotiation as we are requesting a node number no negotiating for one.

src/MinimumNodeService.h
MartinDaCosta53: I agree that the heartbeat can be a bit of a nuisance during development and testing.  Why not just set the default at line 51 bool noHeartbeat = 1; ?

SvenRosvall: I didn't want to change the code whenever I wanted to do some testing. That is why I added this function so I could change it from test code.

src/MinimumNodeService.cpp
```diff
@@ -76,20 +76,9 @@ void MinimumNodeService::setUninitialised()
...
-{
   renegotiating = true;
```
SvenRosvall, line 82: Rename renogitiating to "requestingNewNN"

PR #21, opened 2023-09-02
More fixes to problems highlighted by David Ellis Compliance Tests

MartinDaCosta53:
More fixes to problems highlighted by David Ellis Compliance Tests

### DIFF
--- a/src/EventTeachingService.cpp
+++ b/src/EventTeachingService.cpp
@@ -11,259 +11,426 @@
 
 namespace VLCB {
 
-void EventTeachingService::setController(Controller *cntrl)
+void EventTeachingService::setController(Controller *cntrl) 
 {
   this->controller = cntrl;
   this->module_config = cntrl->module_config;
 }
 
-void EventTeachingService::enableLearn()
+void EventTeachingService::enableLearn() 
 {
   bLearn = true;
-   //DEBUG_SERIAL << F("> set learn mode ok") << endl;
+  DEBUG_SERIAL << F("> set learn mode ok") << endl;
   // set bit 5 in parameter 8
   bitSet(controller->_mparams[8], 5);
 }
 
-void EventTeachingService::inhibitLearn()
+void EventTeachingService::inhibitLearn() 
 {
   bLearn = false;
-   //DEBUG_SERIAL << F("> learn mode off") << endl;
+  DEBUG_SERIAL << F("> learn mode off") << endl;
   // clear bit 5 in parameter 8
   bitClear(controller->_mparams[8], 5);
 }
 
-Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
-{
+Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg) {
   unsigned int nn = (msg->data[1] << 8) + msg->data[2];
   unsigned int en = (msg->data[3] << 8) + msg->data[4];
-  // DEBUG_SERIAL << ">CbusSvc handling message op=" << _HEX(opc) << " nn=" << nn << " en" << en << endl;
+  DEBUG_SERIAL << ">VlcbSvc handling message op=" << _HEX(opc) << " nn=" << nn << " en" << en << endl;
 
-  switch (opc)
-  {
-    
-  case OPC_NNLRN:
-    // received NNLRN -- place into learn mode
-    //DEBUG_SERIAL << F("> NNLRN for node = ") << nn << F(", learn mode on") << endl;
+  switch (opc) {
 
-    if (nn == module_config->nodeNum)
-    {
-     enableLearn();
-    }
+    case OPC_NNLRN:
+      // 53 - place into learn mode
+      //DEBUG_SERIAL << F("> NNLRN for node = ") << nn << F(", learn mode on") << endl;
 
-    return PROCESSED;
-
-  case OPC_EVULN:
-    // received EVULN -- unlearn an event, by event number
-    // en = (msg->data[3] << 8) + msg->data[4];
-    // DEBUG_SERIAL << F("> EVULN for nn = ") << nn << F(", en = ") << en << endl;
-
-    // we must be in learn mode
-    if (bLearn == true)
-    {
-      // DEBUG_SERIAL << F("> searching for existing event to unlearn") << endl;
-
-      // search for this NN and EN pair
-      byte index = module_config->findExistingEvent(nn, en);
-
-      if (index < module_config->EE_MAX_EVENTS)
+      if (nn == module_config->nodeNum) 
       {
-        // DEBUG_SERIAL << F("> deleting event at index = ") << index << F(", evs ") << endl;
-        module_config->cleareventEEPROM(index);
-
-        // update hash table
-        module_config->updateEvHashEntry(index);
-
-        // respond with WRACK
-        controller->sendWRACK();
-      } else {
-        // DEBUG_SERIAL << F("> did not find event to unlearn") << endl;
-        // respond with CMDERR
-        controller->sendCMDERR(10);
+        enableLearn();
+      } 
+      else  //if we are learning and another node is put in learn mode, stop learn.
+      {
+        if (bLearn) {
+          inhibitLearn();
+        }
       }
-    } // if in learn mode
 
-    return PROCESSED;
+      return PROCESSED;
 
-  case OPC_NNULN:
-    // received NNULN -- exit from learn mode
+    case OPC_EVULN:
+      // 95 - unlearn an event, by event number
+      // en = (msg->data[3] << 8) + msg->data[4];
+      // DEBUG_SERIAL << F("> EVULN for nn = ") << nn << F(", en = ") << en << endl;
 
-    if (nn == module_config->nodeNum)
-    {
-      inhibitLearn();
-    }
-
-    return PROCESSED;
-
-  case OPC_RQEVN:
-    // received RQEVN -- request for number of stored events
-    // DEBUG_SERIAL << F("> RQEVN -- number of stored events for nn = ") << nn << endl;
-
-    if (nn == module_config->nodeNum)
-    {
-      // respond with 0x74 NUMEV
-
-      controller->sendMessageWithNN(OPC_NUMEV, module_config->numEvents());
-    }
-
-    return PROCESSED;
-
-  case OPC_NERD:
-    // request for all stored events
-    // DEBUG_SERIAL << F("> NERD : request all stored events for nn = ") << nn << endl;
-
-    if (nn == module_config->nodeNum)
-    {
-      msg->len = 8;
-      msg->data[0] = OPC_ENRSP;                       // response opcode
-      msg->data[1] = highByte(module_config->nodeNum);        // my NN hi
-      msg->data[2] = lowByte(module_config->nodeNum);         // my NN lo
-
-      for (byte i = 0; i < module_config->EE_MAX_EVENTS; i++)
+      // we must be in learn mode
+      if (bLearn == true) 
       {
-        if (module_config->getEvTableEntry(i) != 0) {
-          // it's a valid stored event
+        if (msg->len < 5) 
+        {
+          controller->sendGRSP(OPC_EVULN, getServiceID(), CMDERR_INV_CMD);
+          controller->sendCMDERR(CMDERR_INV_CMD);
 
+        } 
+        else 
+        {
+          // DEBUG_SERIAL << F("> searching for existing event to unlearn") << endl;
+
+          // search for this NN and EN pair
+          byte index = module_config->findExistingEvent(nn, en);
+
+          if (index < module_config->EE_MAX_EVENTS) 
+          {
+            // DEBUG_SERIAL << F("> deleting event at index = ") << index << F(", evs ") << endl;
+            module_config->cleareventEEPROM(index);
+
+            // update hash table
+            module_config->updateEvHashEntry(index);
+
+            // respond with WRACK
+            controller->sendWRACK();
+            controller->sendGRSP(OPC_EVULN, getServiceID(), GRSP_OK);
+          } 
+          else 
+          {
+            // DEBUG_SERIAL << F("> did not find event to unlearn") << endl;
+            // respond with CMDERR
+            controller->sendCMDERR(CMDERR_INV_NV_IDX);
+            controller->sendGRSP(OPC_EVULN, getServiceID(), CMDERR_INV_NV_IDX);
+          }
+        }
+      }  // if in learn mode
+
+      return PROCESSED;
+
+    case OPC_NNULN:
+      // 54 - exit from learn mode
+
+      if (nn == module_config->nodeNum) 
+      {
+        inhibitLearn();
+      }
+
+      return PROCESSED;
+
+    case OPC_RQEVN:
+      // 58 - request for number of stored events
+      // DEBUG_SERIAL << F("> RQEVN -- number of stored events for nn = ") << nn << endl;
+
+      if (nn == module_config->nodeNum) 
+      {
+        // respond with 0x74 NUMEV
+        controller->sendMessageWithNN(OPC_NUMEV, module_config->numEvents());
+      }
+
+      return PROCESSED;
+
+    case OPC_NERD:
+      // 57 - request for all stored events
+      DEBUG_SERIAL << F("> NERD : request all stored events for nn = ") << nn << endl;
+
+      if (nn == module_config->nodeNum) 
+      {
+        msg->len = 8;
+        msg->data[0] = OPC_ENRSP;     // response opcode
+        msg->data[1] = highByte(nn);  // my NN hi
+        msg->data[2] = lowByte(nn);   // my NN lo
+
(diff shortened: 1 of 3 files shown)


src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
   this->controller = cntrl;
   this->module_config = cntrl->module_config;
 }
 
-void EventTeachingService::enableLearn()
+void EventTeachingService::enableLearn() 
 {
   bLearn = true;
-   //DEBUG_SERIAL << F("> set learn mode ok") << endl;
+  DEBUG_SERIAL << F("> set learn mode ok") << endl;
```
SvenRosvall, line 23: Please comment out the debug outputs again once you are done with trouble shooting.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
 
-  case OPC_RQEVN:
-    // received RQEVN -- request for number of stored events
-    // DEBUG_SERIAL << F("> RQEVN -- number of stored events for nn = ") << nn << endl;
+      if (nn == module_config->nodeNum) 
+      {
+        msg->len = 8;
+        msg->data[0] = OPC_ENRSP;     // response opcode
+        msg->data[1] = highByte(nn);  // my NN hi
+        msg->data[2] = lowByte(nn);   // my NN lo
```
SvenRosvall, line 136: Well spotted that you can use "nn" here!

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
-    // DEBUG_SERIAL << F("> NERD : request all stored events for nn = ") << nn << endl;
+      return PROCESSED;
 
-    if (nn == module_config->nodeNum)
-    {
-      msg->len = 8;
-      msg->data[0] = OPC_ENRSP;                       // response opcode
-      msg->data[1] = highByte(module_config->nodeNum);        // my NN hi
-      msg->data[2] = lowByte(module_config->nodeNum);         // my NN lo
+    case OPC_NENRD:
```
SvenRosvall, line 158: What about checking for short messages? You do that everywhere else. ;)

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
 
-      for (byte i = 0; i < module_config->EE_MAX_EVENTS; i++)
+      if (nn == module_config->nodeNum) 
       {
-        if (module_config->getEvTableEntry(i) != 0) {
-          // it's a valid stored event
+        msg->len = 8;
+        msg->data[0] = OPC_ENRSP;     // response opcode
+        msg->data[1] = highByte(nn);  // my NN hi
+        msg->data[2] = lowByte(nn);   // my NN lo
```
SvenRosvall, line 167: These four lines that set the message data should be moved down to line 181 where we are sending this message.
First of all, you only need these data set when we are sending the message in the else branch.
Secondly it is confusing that you are writing to the msg structure and then on line 169 you read from it.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
-        // DEBUG_SERIAL << F("> request for invalid event index") << endl;
-        controller->sendCMDERR(6);
+        if (msg->len < 5) 
+        {
+          controller->sendGRSP(OPC_REVAL, getServiceID(), CMDERR_INV_CMD);
+          controller->sendCMDERR(CMDERR_INV_CMD);
+        } 
+        else 
+        {
+          if (module_config->getEvTableEntry(msg->data[3]) != 0)
```
SvenRosvall, line 203: Should we also check if the event index is too large?

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
-      // not found - it's a new event
-      if (index >= module_config->EE_MAX_EVENTS)
+      // we must be in learn mode
+      if (bLearn) 
       {
-        // DEBUG_SERIAL << F("> existing event not found - creating a new one if space available") << endl;
-        index = module_config->findEventSpace();
+        if (msg->len < 7) 
+        {
+          controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_CMD);
```
SvenRosvall, line 282: Should we send a CMDERR for short messages?

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+          byte index = msg->data[5];
+          byte evIndex = msg->data[6];
+          byte evVal = msg->data[7];
+
+          // invalid index
+          if (index >= module_config->EE_MAX_EVENTS) 
+          {
+            DEBUG_SERIAL << F("> invalid index") << endl;
+            controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_EV_IDX);
+            return PROCESSED;
```
SvenRosvall, line 358: This return is superflous as you have a return at the end of the case-block. 
You don't have a return in the other branches. I don't care which style you use, just want it to be consistent. :)

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+
+      DEBUG_SERIAL << endl
+                   << F("> EVLRNI for source nn = ") << nn << endl;
+
+      // we must be in learn mode
+      if (bLearn) 
+      {
+        if (msg->len < 8) 
+        {
+          controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_CMD);
```
SvenRosvall, line 344: OP-code spec says nothing about GRSP for any of the error cases. I guess because it is deprecated in VLCB.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+          // write the event data
+          else 
+          {
+            // write the event to EEPROM at this location -- EVs are indexed from 1 but storage offsets start at zero !!
+            DEBUG_SERIAL << F("> writing EV = ") << evIndex << F(", at index = ") << index << F(", offset = ") << (module_config->EE_EVENTS_START + (index * module_config->EE_BYTES_PER_EVENT)) << endl;
+
+            // don't repeat this for subsequent EVs
+            if (evIndex < 2) 
+            {
+              module_config->writeEvent(index, &msg->data[1]);
```
SvenRosvall, line 369: This needs more explaining. What are we writing here?  Why are we writing this only for the first event variable?

src/EventTeachingService.h
```diff
@@ -7,29 +7,30 @@
 
 #include "Service.h"
 
-namespace VLCB
-{
+namespace VLCB {
```
SvenRosvall, line 10: Please keep curly brackets "{" on new lines. This is the style we want to use.
Have you set this style in Notepad++? Or is this in Arduino IDE?

src/EventTeachingService.h
```diff
@@ -7,29 +7,30 @@
...
 
-  virtual byte getServiceID() override { return 4; }
-  virtual byte getServiceVersionID() override { return 1; }
-  
+  virtual byte getServiceID() override {
+    return 4;
+  }
+  virtual byte getServiceVersionID() override {
+    return 1;
+  }
```
SvenRosvall, line 24: Prefer to keep these one-liners on a single line when they are in a header file.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
   bLearn = false;
-   //DEBUG_SERIAL << F("> learn mode off") << endl;
+  DEBUG_SERIAL << F("> learn mode off") << endl;
   // clear bit 5 in parameter 8
   bitClear(controller->_mparams[8], 5);
 }
 
-Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
-{
+Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg) {
```
SvenRosvall, line 36: Please keep the curly-brace style were they are on new lines.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
   unsigned int nn = (msg->data[1] << 8) + msg->data[2];
   unsigned int en = (msg->data[3] << 8) + msg->data[4];
-  // DEBUG_SERIAL << ">CbusSvc handling message op=" << _HEX(opc) << " nn=" << nn << " en" << en << endl;
+  DEBUG_SERIAL << ">VlcbSvc handling message op=" << _HEX(opc) << " nn=" << nn << " en" << en << endl;
+
+  switch (opc) {
```
SvenRosvall, line 41: And here.

src/EventTeachingService.cpp
MartinDaCosta53: Sorry, forgot that I'd uncommented some debugs!

src/EventTeachingService.cpp
MartinDaCosta53: The message is only opcode and node number, so if the message is short you don't know if it was for you or not.

src/EventTeachingService.cpp
MartinDaCosta53: Good point.  It's done

src/EventTeachingService.cpp
MartinDaCosta53: This is checked in configuration.cpp which returns 0 if the index is larger than EE_MAX_EVENTS.  So, if it is too large it runs the else code and sends error messages.

src/EventTeachingService.cpp
MartinDaCosta53: Again, this is just the opcode and node number, so if it is short the message is regarded as not for me.

src/EventTeachingService.cpp
MartinDaCosta53: Good point.

src/EventTeachingService.cpp
MartinDaCosta53: I think that the spec should be updated to put in these error messages, particularly as I don't think that this op code should be deprecated.  I have found it useful for teaching short events to producers.  I've added it to the agenda for the next meeting.

src/EventTeachingService.h
MartinDaCosta53: This is the Arduino IDE and it's a pain!  If I use its auto format, I have to go through and correct the misplaced opening brace and sometimes I miss them.  I try to get the formatting all done in Notepad++ now.

src/EventTeachingService.h
MartinDaCosta53: Autoformat, again!

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
-    if (nn == module_config->nodeNum)
-    {
-      inhibitLearn();
-    }
+      return PROCESSED;
 
-    return PROCESSED;
+    case OPC_NERD:
+      // 57 - request for all stored events
+      DEBUG_SERIAL << F("> NERD : request all stored events for nn = ") << nn << endl;
```
SvenRosvall, line 131: One more debug here.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+          {
+            // it's a valid stored event
+            // read the event data from EEPROM
+            // construct and send a ENRSP message
+            module_config->readEvent(i, &msg->data[3]);
+            msg->data[7] = i;  // event table index
 
-      controller->sendMessageWithNN(OPC_NUMEV, module_config->numEvents());
-    }
+            DEBUG_SERIAL << F("> sending ENRSP reply for event index = ") << i << endl;
```
SvenRosvall, line 150: And here.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
-        // write the event to EEPROM at this location -- EVs are indexed from 1 but storage offsets start at zero !!
-        // DEBUG_SERIAL << F("> writing EV = ") << evindex << F(", at index = ") << index << F(", offset = ") << (module_config->EE_EVENTS_START + (index * module_config->EE_BYTES_PER_EVENT)) << endl;
+      return PROCESSED;
+
+    case OPC_EVLRNI:
+      // F5 - learn an event using event index
 
-        // don't repeat this for subsequent EVs
-        if (evindex < 2)
+      DEBUG_SERIAL << endl
```
SvenRosvall, line 337: Another debug output to comment out.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+        else 
+        {
+          byte index = msg->data[5];
+          byte evIndex = msg->data[6];
+          byte evVal = msg->data[7];
+
+          // invalid index
+          if (index >= module_config->EE_MAX_EVENTS) 
+          {
+            DEBUG_SERIAL << F("> invalid index") << endl;
```
SvenRosvall, line 357: More debug outputs.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+          controller->sendCMDERR(CMDERR_INV_CMD);
+          return PROCESSED;
+        }          
+        
+        byte index = module_config->findExistingEvent(nn, en);
+        byte evIndex = msg->data[5];
+
+        if (index >= module_config->EE_MAX_EVENTS) 
+        {
+           DEBUG_SERIAL << F("> event not found") << endl;
```
SvenRosvall, line 403: MOre debug outputs.

src/EventTeachingService.cpp
SvenRosvall: OK.

src/EventTeachingService.cpp
MartinDaCosta53: I think that I've now got them all.

src/EventTeachingService.cpp
MartinDaCosta53: I think that I've now got them all.

src/EventTeachingService.cpp
MartinDaCosta53: I think that I've now got them all.

src/EventTeachingService.h
```diff
@@ -7,29 +7,27 @@
...
 
-namespace VLCB
+namespace VLCB 
 {
 
 class Configuration;
 
-class EventTeachingService : public Service
-{
+class EventTeachingService : public Service {
```
SvenRosvall, line 15: Please move this curly bracket to a new line.

src/EventTeachingService.cpp
```diff
@@ -11,259 +11,427 @@
...
+      //DEBUG_SERIAL << F("> NNLRN for node = ") << nn << F(", learn mode on") << endl;
 
-    return PROCESSED;
+      if (nn == module_config->nodeNum) 
+      {
+        enableLearn();
+      } 
+      else  //if we are learning and another node is put in learn mode, stop learn.
+      {
+        if (bLearn) {
```
SvenRosvall, line 55: One more misplaced curly bracket.

src/EventTeachingService.h
MartinDaCosta53: Sorry, missed that one!

src/EventTeachingService.cpp
MartinDaCosta53: Got it

src/EventTeachingService.cpp
MartinDaCosta53: The event will be there because otherwise it won't have an index.  I am using this opcode to teach producers short events.  It can also be used to allow a producer to spoof.  I suppose the other way would be to check if the NN and the EN in the event table match the incoming for all instances of the op code and change the table if not.  I'll look at this as possible enhancement.

MartinDaCosta53: Now revised how NN & EN are checked and updated as necessary.  This code compiles but I haven't been able to fully test it yet.  Is there any reason why this shouldn't run OK?

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
SvenRosvall: No, "if (eventTableNNEN != &msg->data[1])" won't work. This is comparing addresses to the data. To compare the four bytes you should write this as "if (memcmp(eventTableNNEN, &msg->data[1], 4) == 0)"
"memcpy" works similar to "strcmp" that I covered in the talk about strings, but it works off memory so doesn't check for terminating null byte. Instead you need the "4" as a length of memory to be compared.

MartinDaCosta53: Ah, thanks Sven.  Thought it looked too simple!  Now corrected

MartinDaCosta53:
Standardisation of formatting and code enhancements.

PR #23, opened 2023-09-05
Introduce Event Producer Service

MartinDaCosta53:
There are some significant change to the example VLCB_1in1out.  Happy to explain how this all works on Zoom, if you like

### DIFF
--- a/examples/VLCB_1in1out/VLCB_1in1out.ino
+++ b/examples/VLCB_1in1out/VLCB_1in1out.ino
@@ -208,7 +208,7 @@ void processModuleSwitchChange()
 {
   if (moduleSwitch.stateChanged())
   {
-    bool state = moduleSwitch.isPressed() ? true : false;
+    bool state = moduleSwitch.isPressed();
     byte eventNumber = 1;
     epService.sendEvent(state, eventNumber);
   }
--- a/examples/VLCB_1in1out/VLCB_1in1out.ino
+++ b/examples/VLCB_1in1out/VLCB_1in1out.ino
@@ -26,6 +26,7 @@
 #include "CanService.h"
 #include "NodeVariableService.h"
 #include "EventConsumerService.h"
+#include "EventProducerService.h"
 #include "EventTeachingService.h"
 
 // constants
@@ -50,8 +51,9 @@ VLCB::CanService canService;
 VLCB::NodeVariableService nvService;
 VLCB::EventConsumerService ecService;
 VLCB::EventTeachingService etService;
+VLCB::EventProducerService epService;
 VLCB::Controller controller(&userInterface, &modconfig, &can2515, 
-                            { &mnService, &canService, &nvService, &ecService, &etService }); // Controller object
+                            { &mnService, &canService, &nvService, &ecService, &epService, &etService }); // Controller object
 
 // module objects
 VLCB::Switch moduleSwitch(5);            // an example switch as input
@@ -76,8 +78,9 @@ void setupVLCB()
   modconfig.EE_NUM_NVS = 10;
   modconfig.EE_EVENTS_START = 20;
   modconfig.EE_MAX_EVENTS = 32;
+  modconfig.EE_PROD_EVENTS = 1;
   modconfig.EE_NUM_EVS = 1;
-  modconfig.EE_BYTES_PER_EVENT = (modconfig.EE_NUM_EVS + 4);
+ 
 
   // initialise and load configuration
   modconfig.begin();
@@ -205,17 +208,9 @@ void processModuleSwitchChange()
 {
   if (moduleSwitch.stateChanged())
   {
-    byte opc = moduleSwitch.isPressed() ? OPC_ACON : OPC_ACOF;
+    bool state = moduleSwitch.isPressed() ? true : false;
     byte eventNumber = 1;
-
-    if (controller.sendMessageWithNN(opc, 0, eventNumber))
-    {
-      Serial << F("> sent VLCB message") << endl;
-    }
-    else
-    {
-      Serial << F("> error sending VLCB message") << endl;
-    }
+    epService.sendEvent(state, eventNumber);
   }
 }
 
--- a/src/Controller.h
+++ b/src/Controller.h
@@ -70,8 +70,7 @@ public:
   void indicateActivity();
   void setLearnMode(byte reqMode);
   void setFrameHandler(void (*fptr)(CANFrame *msg), byte *opcodes = NULL, byte num_opcodes = 0);
-  byte newMode;
-
+  
 private:                                          // protected members become private in derived classes
   UserInterface *_ui;
   Configuration *module_config;
@@ -83,6 +82,7 @@ private:                                          // protected members become pr
   void (*framehandler)(CANFrame *msg) = NULL;
   byte *_opcodes = NULL;
   byte _num_opcodes = 0;
+  byte newMode;
 
   bool filterByOpcodes(const CANFrame *msg) const;
   void callFrameHandler(CANFrame *msg);
--- a/src/Controller.h
+++ b/src/Controller.h
@@ -70,6 +70,7 @@ public:
   void indicateActivity();
   void setLearnMode(byte reqMode);
   void setFrameHandler(void (*fptr)(CANFrame *msg), byte *opcodes = NULL, byte num_opcodes = 0);
+  byte newMode;
 
 private:                                          // protected members become private in derived classes
   UserInterface *_ui;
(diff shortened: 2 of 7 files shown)


examples/VLCB_1in1out/VLCB_1in1out.ino
```diff
@@ -205,17 +208,9 @@ void processModuleSwitchChange()
 {
   if (moduleSwitch.stateChanged())
   {
-    byte opc = moduleSwitch.isPressed() ? OPC_ACON : OPC_ACOF;
+    bool state = moduleSwitch.isPressed() ? true : false;
```
SvenRosvall, line 211: You can remove the "? true : false" part as .isPressed() returns the boolean you want to set state to.

src/EventProducerService.h
```diff
@@ -0,0 +1,45 @@
...
+  void sendEvent(bool state, byte index, byte data1, byte data2);
+  void sendEvent(bool state, byte index, byte data1, byte data2, byte data3);
+
+private:
+  Controller *controller;
+  Configuration *module_config;  // Shortcut to reduce indirection code.
+  void (*eventhandler)(byte index, CANFrame *msg);
+  void (*eventhandlerex)(byte index, CANFrame *msg, bool evOn, byte evVal);
+
+  byte instantMode;
```
SvenRosvall, line 40: It is confusing that you have an "instantMode" here and in MinimumNodeService. Do they do the same thing or just similar?

src/EventProducerService.cpp
```diff
@@ -0,0 +1,217 @@
...
+  }
+  else
+  {
+    setProducedEvents();
+  }
+}
+
+void EventProducerService::setProducedEvents()
+{
+  if (module_config->EE_PROD_EVENTS > 0)
```
SvenRosvall, line 41: This if statement is redundant. The for loop won't execute if EE_PROD_EVENTS == 0.

src/EventProducerService.cpp
```diff
@@ -0,0 +1,217 @@
...
+  {
+    return;
+  }
+  else
+  {
+    setProducedEvents();
+  }
+}
+
+void EventProducerService::setProducedEvents()
```
SvenRosvall, line 39: Is this setting up own consumed events?
In that case I would rename it setupOwnEvents().
Also you could check the parameter flags with PF_COE. If the module hasn't set this then there is no need to fill this table.

src/EventProducerService.h
MartinDaCosta53: They are similar in that they both note the mode at that instant to allow the service to determine where it is going.  Could change the instance in here to presentMode, if you like.

src/EventProducerService.cpp
```diff
@@ -0,0 +1,217 @@
...
+        // Event does not exist so create it
+        byte data[4];
+        data[0] = highByte(module_config->nodeNum);
+        data[1] = lowByte(module_config->nodeNum);
+        data[2] = 0;
+        data[3] = i;
+        
+        module_config->writeEvent(i, data);  //index = i
+        module_config->writeEventEV(i, 1, 1);
+        module_config->writeEventEV(i, 2, 0);  //set default switch type
```
SvenRosvall, line 56: Please add a comment to describe what EV1 and EV2 do.
What if the user sketch has set EE_NUM_EVS < 2 ?

src/EventProducerService.cpp
MartinDaCosta53: Oh yes. You're right!  I'll change it

src/EventProducerService.cpp
MartinDaCosta53: No. Produced events are now in the events table so that they can be taught.  I haven't done consume own events yet.

src/EventProducerService.cpp
```diff
@@ -0,0 +1,217 @@
...
+void EventProducerService::process(byte num)
+{
+  // Do this if mode changes
+  instantMode = controller->newMode;
+  if (lastMode != instantMode)
+  {
+    // Do this if change from setup to normal.
+    if ((lastMode == MODE_SETUP) && (instantMode == MODE_NORMAL))
+    {
+      setProducedEvents();
```
SvenRosvall, line 72: Isn't it enough to call setProducedEvents() once, i.e. from begin().

src/EventProducerService.cpp
MartinDaCosta53: Comment added
There needs to be a trap for such an error.  Need to ad to the TODO list.

src/EventProducerService.h
```diff
@@ -0,0 +1,45 @@
...
+  void sendEvent(bool state, byte index);
+  void sendEvent(bool state, byte index, byte data1);
+  void sendEvent(bool state, byte index, byte data1, byte data2);
+  void sendEvent(bool state, byte index, byte data1, byte data2, byte data3);
+
+private:
+  Controller *controller;
+  Configuration *module_config;  // Shortcut to reduce indirection code.
+  void (*eventhandler)(byte index, CANFrame *msg);
+  void (*eventhandlerex)(byte index, CANFrame *msg, bool evOn, byte evVal);
```
SvenRosvall, line 38: "eventhandlerex" is not used anywhere.

src/EventProducerService.h
```diff
@@ -0,0 +1,45 @@
...
+  void begin();
+  void sendEvent(bool state, byte index);
+  void sendEvent(bool state, byte index, byte data1);
+  void sendEvent(bool state, byte index, byte data1, byte data2);
+  void sendEvent(bool state, byte index, byte data1, byte data2, byte data3);
+
+private:
+  Controller *controller;
+  Configuration *module_config;  // Shortcut to reduce indirection code.
+  void (*eventhandler)(byte index, CANFrame *msg);
```
SvenRosvall, line 37: You don't have a setter for "eventhandler" that the user can call to set up this handler.

src/EventProducerService.cpp
MartinDaCosta53: No, because if there has been a reset to manufacturers defaults (OPC_NNRSM) the vents table will have been cleared and the default produced evets will need to be re-written.

src/EventProducerService.cpp
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }    
+  }
+}
+
+void EventProducerService::process(byte num)
+{
+  // Do this if mode changes
+  instantMode = controller->newMode;
+  if (lastMode != instantMode)
```
SvenRosvall, line 67: "lastMode" might not be initialized here. The begin() function would be a good place to set lastMode.

src/EventProducerService.h
MartinDaCosta53: No, it isn't.  I haven't removed it yet because I didn't know what I might need for AREQ.

src/EventProducerService.cpp
MartinDaCosta53: Done

examples/VLCB_1in1out/VLCB_1in1out.ino
MartinDaCosta53: Of course it does!  Silly me!

src/EventProducerService.h
MartinDaCosta53: I think that I'll remove "eventhandler" and "eventhandlerex" until I know if I'll need them.

src/Controller.h
```diff
@@ -70,6 +70,7 @@ class Controller
   void indicateActivity();
   void setLearnMode(byte reqMode);
   void setFrameHandler(void (*fptr)(CANFrame *msg), byte *opcodes = NULL, byte num_opcodes = 0);
+  byte newMode;
```
SvenRosvall, line 73: This variable can be placed in the private part. EventProducerService is a "friend" which means it has access to everything in Controller.

src/EventProducerService.h
MartinDaCosta53: I've removed it

src/EventProducerService.cpp
SvenRosvall: Ah.

src/Controller.h
MartinDaCosta53: I did wonder.  Now moved.

src/EventProducerService.cpp
SvenRosvall: Interesting edge case. So if I understand you there is no need to do this every time we leave setup mode. Possibly there should be another mechanism where things are set up after a reset (or from Uninitialized.)

src/EventProducerService.h
SvenRosvall: No. Just confusing to have two similar names. So keep it as "instantMode" if it is needed.

src/EventProducerService.cpp
MartinDaCosta53: Actually, we don't need it!  After a resetModule() we reboot()! (Configuration.cpp line 567 - 597).  I'll take it out.

MartinDaCosta53: That leaves process() empty so I'll lose that, as well.

PR #24, opened 2023-09-05
Remove all mode variables from EventProducerService

MartinDaCosta53:


src/Controller.cpp
```diff
@@ -117,11 +117,20 @@ bool Controller::isRTR(CANFrame *amsg)
 
 void Controller::indicateMode(byte mode)
 {
-  newMode = mode;  // used by Event Producer Service
   // DEBUG_SERIAL << F("> indicating mode = ") << mode << endl;
   if (_ui) {
     _ui->indicateMode(mode);
   }
+  
+  if (mode == MODE_NORMAL) // used by Event Producer Service
```
SvenRosvall, line 125: I wonder if this is a bit blunt. Isn't this condition true for any transition to NORMAL?

src/EventProducerService.cpp
```diff
@@ -22,13 +22,8 @@ void EventProducerService::setController(Controller *cntrl)
...
 void EventProducerService::begin()
-{
-  //Initialise instantMode
-  instantMode = module_config->currentMode;
-  lastMode = instantMode;
-  //DEBUG_SERIAL << F("> instant MODE initialise as: ") << instantMode << endl;
-  
-  if (instantMode == MODE_UNINITIALISED)
+{  
+  if (module_config->currentMode == MODE_UNINITIALISED)
```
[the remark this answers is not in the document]
MartinDaCosta53: Done and other branch removed as superfluous

src/Controller.cpp
MartinDaCosta53: The route by which it is called only occurs if there has been a transition from MODE_SETUP to MODE_NORMAL.  coming to or from LEARN or NHEARTBEAT does not cause a change to indicate mode.  Thus, it is only if we have requested a node number or returned to the old one on timeout that it is called, which is just what we want.

PR #28, opened 2023-09-08
Set params[8] flags accordong to enrolled services and operating mode

MartinDaCosta53:


src/Controller.cpp
```diff
@@ -84,7 +84,30 @@ void Controller::setFrameHandler(void (*fptr)(CANFrame *msg), byte opcodes[], by
...
+  bool consumer = 0;
+  bool nPreNormal = 0;
+  bool conOwn = 0;
+  byte mode = module_config->currentMode;
+  
   _mparams = mparams;
+  
+  for (Service * svc : services)
+  {
+    if (svc->getServiceID() == 5)
```
SvenRosvall, line 97: Lots of anonymous constants and bit fiddling here.
What is serviceID 5 here? Use SERVICE_ID_PRODUCER.
See separate email on how to rewrite this function.

src/Controller.cpp
```diff
@@ -122,6 +145,15 @@ void Controller::indicateMode(byte mode)
     _ui->indicateMode(mode);
   }
   
+  if ((mode == MODE_UNINITIALISED) || (mode == MODE_SETUP))
+  {
+    bitClear(_mparams[8], 2);
```
SvenRosvall, line 150: I would use PF_FLIM constant here instead of a literal '2'. 
(Yes, I will add a PF_NORMAL to replace PF_FLIM.)
Then replace the bitClear with "_params[8] &= PF_FLIM;

src/Controller.cpp
```diff
@@ -122,6 +145,15 @@ void Controller::indicateMode(byte mode)
...
+  }
+  else
+  {
+    bitSet(_mparams[8], 2);
```
SvenRosvall, line 154: And similarly:
_params[8] |= PF_FLIM;

src/Controller.cpp
```diff
@@ -84,7 +84,30 @@ void Controller::setFrameHandler(void (*fptr)(CANFrame *msg), byte opcodes[], by
 //
 void Controller::setParams(unsigned char *mparams)
 {
+  bool producer = 0;
+  bool consumer = 0;
+  bool nPreNormal = 0;
+  bool conOwn = 0;
```
SvenRosvall, line 90: You never set conOwn here.
We talked about a COEService at some stage. Would this be a good idea? It may be responsible for putting produced events into the events table.

src/Controller.cpp
```diff
@@ -84,7 +84,30 @@ void Controller::setFrameHandler(void (*fptr)(CANFrame *msg), byte opcodes[], by
...
+    if (svc->getServiceID() == 5)
+    {
+      producer = 1;
+    }
+    if (svc->getServiceID() == 6)
+    {
+      consumer = 1;
+    }
+  }
+  if ((mode != MODE_UNINITIALISED) && (mode != MODE_SETUP))
```
SvenRosvall, line 106: The persisted currentMode cannot be MODE_SETUP, right?

MartinDaCosta53: Correct!  This can be much simplified

src/Controller.cpp
MartinDaCosta53: I put it there as a place holder because it is a flag in params[8].  However, we know we'll need to come back to it when we decide how COE will be done so I'll leave it out, for now.

src/Controller.cpp
MartinDaCosta53: Yes, much better.  It then gets rid of the Arduino macros

src/Controller.cpp
```diff
@@ -122,6 +141,15 @@ void Controller::indicateMode(byte mode)
     _ui->indicateMode(mode);
   }
   
+  if ((mode == MODE_UNINITIALISED) || (mode == MODE_SETUP))
+  {
+    _mparams[PAR_FLAGS] &= ~PF_FLiM;
```
SvenRosvall, line 146: Sorry for giving you the wrong code example. Of course there has to be a ~ here.

PR #29, opened 2023-09-08
OPC_RQNPN brought in line with VLCB specification

MartinDaCosta53:


src/MinimumNodeService.cpp
```diff
@@ -486,11 +485,4 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
   }
 }
 
-void MinimumNodeService::setSetupMode()
-{
-  bModeSetup = true;
-  instantMode = MODE_SETUP;
-  timeOutTimer = 0;
-}
-
```
SvenRosvall, line 495: Can you bring this back in? 
Seems like you have overwritten an update from me.

src/MinimumNodeService.cpp
```diff
@@ -209,14 +200,22 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
...
-          if (paran <= controller->_mparams[0])
+           //DEBUG_SERIAL << F("> RQNPN request for parameter # ") << paran << F(", from nn = ") << nn << endl;
+          
+          if (paran == 0)
+          {
+            for (byte i = 0; i <= controller->_mparams[0]; i++)
+            {
+              controller->sendMessageWithNN(OPC_PARAN, i, controller->_mparams[i]);
+              delay(1);
```
SvenRosvall, line 210: I have put this delay(1) in my local code in the CanTransport class. This is cleaner and fixes the same problem for all places where we send many messages. Maybe I should have checked in that?
I didn't check it in as I wanted to do a proper fix where we check the transmit queue size.

src/MinimumNodeService.cpp
MartinDaCosta53: Sorry about that.  Now back in

src/MinimumNodeService.cpp
MartinDaCosta53: Delay removed

MartinDaCosta53: I found that I needed the delay because the FCU can also be trying to request a sequence of PARANs.  Ian & I had a discussion in the OpCodes spec on Google Docs

PR #31, opened 2023-09-09
Corrections from test software

MartinDaCosta53:


src/EventTeachingService.cpp
```diff
@@ -22,15 +22,17 @@ void EventTeachingService::enableLearn()
...
 }
 
 void EventTeachingService::inhibitLearn() 
 {
   bLearn = false;
   //DEBUG_SERIAL << F("> learn mode off") << endl;
   // clear bit 5 in parameter 8
-  bitClear(controller->_mparams[8], 5);
+  controller->_mparams[PAR_FLAGS] &= ~PF_LRN;
+ // bitClear(controller->_mparams[8], 5);
```
SvenRosvall, line 35: Happy to remove the commented out bit functions.

src/EventTeachingService.cpp
```diff
@@ -150,7 +152,7 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
 
             //DEBUG_SERIAL << F("> sending ENRSP reply for event index = ") << i << endl;
             controller->sendMessage(msg);
-            delay(10);
+ //           delay(10);
```
SvenRosvall, line 155: If it works without this delay, please remove this line. No need for such commented code.

src/EventTeachingService.cpp
```diff
@@ -419,7 +433,7 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
             for (byte i = 1; i <= module_config->EE_NUM_EVS; i++)
             {
               controller->sendMessageWithNN(OPC_EVANS, i, module_config->getEventEVval(index, i));
-              delay(10);
+//              delay(10);
```
SvenRosvall, line 436: If it works without this delay, please remove this line. No need for such commented code.

src/EventTeachingService.cpp
```diff
@@ -407,6 +415,12 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
           controller->sendCMDERR(CMDERR_INVALID_EVENT);
           return PROCESSED;
         }
+        
+        if (evindex > module_config->EE_NUM_EVS)
+        { controller->sendCMDERR(CMDERR_INV_EV_IDX);
```
SvenRosvall, line 420: New line please after the curly bracket.

src/MinimumNodeService.cpp
```diff
@@ -82,6 +82,8 @@ void MinimumNodeService::initSetupFromNormal()
   // DEBUG_SERIAL << F("> reverting to Uninitialised mode") << endl;
   requestingNewNN = true;
   controller->sendMessageWithNN(OPC_NNREL);
+  controller->sendGRSP(OPC_MODE, getServiceID(), GRSP_OK);
```
SvenRosvall, line 85: You already have a sendGRSP(OK) where this is called from on line 402, 445. 
I prefer to have the response messages sent in the part of the code where they are called from. 
If there is a problem with sequencing of response messages, then put the sendGRSP() before the call to initSetupFromNormal().

SvenRosvall, line 85: One side effect of this change is that you get a GRPS(OK) also when you press the push button. Not sure this is what you want.

src/EventTeachingService.cpp
```diff
@@ -22,15 +22,17 @@ void EventTeachingService::enableLearn()
...
+ // bitSet(controller->_mparams[8], 5);
 }
 
 void EventTeachingService::inhibitLearn() 
 {
   bLearn = false;
   //DEBUG_SERIAL << F("> learn mode off") << endl;
   // clear bit 5 in parameter 8
-  bitClear(controller->_mparams[8], 5);
+  controller->_mparams[PAR_FLAGS] &= ~PF_LRN;
```
SvenRosvall, line 34: Good to see the constants used instead of numbers.

src/MinimumNodeService.cpp
MartinDaCosta53: OK. It makes sense.  Done.

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/EventTeachingService.cpp
MartinDaCosta53: Done

src/MinimumNodeService.cpp
MartinDaCosta53: Hadn't thought of that! Now sorted

src/EventTeachingService.cpp
MartinDaCosta53: Gives consistency with flag handling elsewhere and clearer

src/EventTeachingService.cpp
dpharris: Should the various delays have names?  Might make them easier to modify at a later date.

MartinDaCosta53: @dpharris The delay was only there so as not to overload receiving modules with a mini storm of messages. It is now unnecessary here because CAN2515 puts a delay in between sending consecutive messages.

src/MinimumNodeService.cpp
SvenRosvall: That is why we do code reviews. :)

MartinDaCosta53: Yes. And I'm pleased to be a part of something that is working well!

PR #32, opened 2023-09-09
Replace literals

SvenRosvall:
This is to make use of the constants in VLCB-defs for self-documentation and for avoiding errors.

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -233,12 +233,19 @@ void Configuration::readEvent(byte idx, byte tarr[])
   // DEBUG_SERIAL << F("> readEvent - idx = ") << idx << F(", nn = ") << (tarr[0] << 8) + tarr[1] << F(", en = ") << (tarr[2] << 8) + tarr[3] << endl;
 }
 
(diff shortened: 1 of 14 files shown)


src/Configuration.cpp
MartinDaCosta53, line 1: Where is the value of EE_HASH_BYTES defined?  It is derived from the number of Node Number bytes plus the number of Event Number bytes and so is always 4.  That is why I took it out of the user sketch and placed it in the library.  No issue with the constant but it needs to be defined in Configuration.h in my view

SvenRosvall: It is defined in Configuration.h. See line 17.

SvenRosvall: Just a note that it was EE_BYTES_PER_EVENT you took out. This is 4 bytes plus the number of event variables.

MartinDaCosta53: Ah yes.  I didn't see the associated change in Configuration.h.  Now I've found it.
Yes, I agree about EE_BYTES_PER_EVENT, but I think that the reason for the the constant being 4 in both cases is the same.

PR #36, opened 2023-09-21
Changes to opcode MODE operation (radical)

MartinDaCosta53:
Radical changes to opcode MODE implementation in line with Ian Hogg's proposal modified by Martin Da Costa's amendments.  If MNS has validated the NN, is in NORMAL mode, but does not respond to the particular command, the message is passed to the next service and so on down the line.  Any service that responds to a command will respond with PROCESSED.
enum ModuleMode{} has been reduced to 3 entries as they are the only ones required. (configuration.h)
In configuration.cpp, EEPROM(0) uses bit 0 to define Uninitialised or Normal and bit 1 to define heartbeat.  heartbeat will always revert to default when setModuleMode sets Normal.  This will happen following a node number change even if mode only goes Normal - Setup - Normal.  heartbeat currently defaults to noHeartbeat, which is at variance with the spec as written but can be easily changed.

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -51,7 +51,7 @@ void Configuration::begin()
   storage->begin();
   loadNVs();
   
-  if ((currentMode == 0xFF) && (nodeNum == 0xFFFF))   // EEPROM is in factory virgin state
+  if ((storage->readEEPROM(0) == 0xFF) && (nodeNum == 0xFFFF))   // EEPROM is in factory virgin state
   {
    resetModule();
    clearResetFlag();
@@ -64,7 +64,23 @@ void Configuration::begin()
 void Configuration::setModuleMode(ModuleMode f)
 {
   currentMode = f;
+  byte oldMode = storage->readEEPROM(0);
   storage->writeEEPROM(0, f);
+}
+
+void Configuration::setHeartbeat(bool beat)
+{
+  heartbeat = beat;
+  byte mode = storage->readEEPROM(0);
+  if (beat)
+  {
+    bitSet(mode, 1);
+  }
+  else
+  {
+    bitClear(mode,1);
+  }
+  storage->writeEEPROM(0, mode);
 }
 
 //
@@ -583,7 +599,7 @@ void Configuration::resetModule()
   // DEBUG_SERIAL << F("> setting Uninitialised config") << endl;
 
   // set the node identity defaults
-  // we set a NN and CANID of zero in Uninitialised as we're now a consumer-only node
+  // we set a NN and CANID of zero and a mode Uninitialised
 
   storage->writeEEPROM(0, 0);     // Mode = Uninitialised
   storage->writeEEPROM(1, 0);     // CANID
@@ -608,7 +624,8 @@ void Configuration::resetModule()
 //
 void Configuration::loadNVs()
 {
-  currentMode = (ModuleMode) storage->readEEPROM(0);
+  currentMode = (ModuleMode) (storage->readEEPROM(0) & 0x01); // Bit 0 persists Uninitialised / Normal mode
+  heartbeat = storage->readEEPROM(0) & 0x02; // Bit 1 persists no heartbeat or heartbaet
   CANID =    storage->readEEPROM(1);
   nodeNum =  (storage->readEEPROM(2) << 8) + storage->readEEPROM(3);
 }
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -70,14 +70,14 @@ void Configuration::setModuleMode(ModuleMode f)
 void Configuration::setHeartbeat(bool beat)
 {
   heartbeat = beat;
-  byte mode = storage->readEEPROM(0);
+  byte mode = storage->readEEPROM(4);
   if (beat)
   {
-    bitSet(mode, 1);
+    bitSet(mode, 0);
   }
   else
   {
-    bitClear(mode,1);
+    bitClear(mode, 0);
   }
   storage->writeEEPROM(0, mode);
 }
@@ -88,11 +88,11 @@ void Configuration::setEventAck(bool ea)
   byte servicePersist = storage->readEEPROM(4);
   if (ea)
   {
-    bitSet(servicePersist, 0);
+    bitSet(servicePersist, 1);
   }
   else
   {
-    bitClear(servicePersist,0);
+    bitClear(servicePersist, 1);
   }
   storage->writeEEPROM(4, servicePersist);
 }
@@ -639,10 +639,10 @@ void Configuration::resetModule()
 void Configuration::loadNVs()
 {
   currentMode = (ModuleMode) (storage->readEEPROM(0) & 0x01); // Bit 0 persists Uninitialised / Normal mode
-  heartbeat = storage->readEEPROM(0) & 0x02; // Bit 1 persists no heartbeat or heartbeat
+  heartbeat = storage->readEEPROM(4) & 0x01; // Bit 1 persists no heartbeat or heartbeat
   CANID =    storage->readEEPROM(1);
   nodeNum =  (storage->readEEPROM(2) << 8) + storage->readEEPROM(3);
-  eventAck = storage->readEEPROM(4) & 0x01; // Bit 0 persists event Acknowledgement if set
+  eventAck = storage->readEEPROM(4) & 0x02; // Bit 0 persists event Acknowledgement if set
 }
 
 //
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -64,7 +64,6 @@ void Configuration::begin()
 void Configuration::setModuleMode(ModuleMode f)
 {
   currentMode = f;
-  byte oldMode = storage->readEEPROM(0);
   storage->writeEEPROM(0, f);
 }
 
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -82,6 +82,21 @@ void Configuration::setHeartbeat(bool beat)
   storage->writeEEPROM(0, mode);
 }
 
+void Configuration::setEventAck(bool ea)
+{
+  eventAck = ea;
+  byte servicePersist = storage->readEEPROM(4);
+  if (ea)
+  {
+    bitSet(servicePersist, 0);
+  }
+  else
+  {
+    bitClear(servicePersist,0);
+  }
+  storage->writeEEPROM(4, servicePersist);
+}
+
 //
 /// store the CANID
 //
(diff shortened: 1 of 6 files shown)


src/Configuration.cpp
```diff
@@ -51,7 +51,7 @@ void Configuration::begin()
   storage->begin();
   loadNVs();
   
-  if ((currentMode == 0xFF) && (nodeNum == 0xFFFF))   // EEPROM is in factory virgin state
+  if ((storage->readEEPROM(0) == 0xFF) && (nodeNum == 0xFFFF))   // EEPROM is in factory virgin state
```
SvenRosvall, line 54: I am OK with using currentMode directly as it is set in loadNVs().

src/Configuration.cpp
```diff
@@ -64,9 +64,25 @@ void Configuration::begin()
 void Configuration::setModuleMode(ModuleMode f)
 {
   currentMode = f;
+  byte oldMode = storage->readEEPROM(0);
```
SvenRosvall, line 67: oldMode is not used anywhere.

src/Configuration.cpp
```diff
@@ -64,9 +64,25 @@ void Configuration::begin()
...
   storage->writeEEPROM(0, f);
```
SvenRosvall, line 68: This overwrites any flags like heartbeat.

src/Configuration.h
SvenRosvall, line 1: I would like some documentation on what each bit field is used for. This can be done with some constants that is used for bit fiddling.
Also I wonder if we should leave more space for future changes. So keep currentMode in eeprom[0] and use eeprom[4] for flags. Node variables start typically from eeprom[20].

src/Configuration.cpp
```diff
@@ -64,9 +64,25 @@ void Configuration::begin()
...
 }
 
+void Configuration::setHeartbeat(bool beat)
```
SvenRosvall, line 71: So this is implementing a "heartbeat on" flag rather than NOHEARTB as we have been discussed. The important part here is what the default is.

src/Configuration.h
```diff
@@ -62,6 +60,7 @@ class Configuration
 
   void setCANID(byte canid);
   void setModuleMode(ModuleMode m);
+  void setHeartbeat(bool beat);
```
SvenRosvall, line 63: Should we have a similar function for event ack? Or is this something that will come later?

src/MinimumNodeService.h
```diff
@@ -51,7 +51,7 @@ class MinimumNodeService : public Service
   
   unsigned long lastHeartbeat = 0;
   byte heartbeatSequence = 0;
-  bool noHeartbeat = false;
+  bool noHeartbeat;
```
SvenRosvall, line 54: Please keep the default initialisation. This reduces problems. Yes I know it will be set at startup but I prefer belt and braces.

src/MinimumNodeService.cpp
```diff
@@ -383,104 +384,43 @@ Processed MinimumNodeService::handleMessage(unsigned int opc, CANFrame *msg)
...
+        }
+        else
+        {
           return PROCESSED;
         }
       }
-      
+      else
+      {
       return PROCESSED;
```
SvenRosvall, line 422: Thinking about these return statements. I don't think we should stop processing here. Some other service might want to do something for this MODE code.

SvenRosvall, line 422: If you want to keep it, please indent this return statement correctly.

src/EventTeachingService.cpp
```diff
@@ -39,6 +39,24 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
 
   switch (opc) 
   {
+    case OPC_MODE:
+      // 76 - Set Operating Mode
+      //DEBUG_SERIAL << F("> MODE -- request op-code received for NN = ") << nn << endl;
+      byte requestedMode = msg->data[3];
```
SvenRosvall, line 45: request mode is visible in the whole switch statement. You can put the code for case OPC_MODE in {...}. 
But looking at other cases, shouldn't there be a test for NORMAL mode? And that we have a node number?
You can only change to or from LEARN mode if you are in NOMRAL mode and have an assigned node number.
If you add such an if-statement then you automatically get a {...} block that limits the visibility of requestedMode.

src/Configuration.cpp
MartinDaCosta53: Thank you! I missed this.  I had it in as a stepping stone as I worked through the changes.  It then became redundant and whilst I took out all of its use, I forgot to take it out!

src/Configuration.h
MartinDaCosta53: I agree about some documentation.  Something that reminds us in the future why we are here helps with looking at maintenance and growth.  I'll have a think about some words in my down time.
I was looking at eeprom(0) as the place for persisted flags as opposed to the bytes used for CAN ID and node number. We only have two at present (Uninititiated / Normal and Haertbeat / NoHeartbeat)  and it seems that there will probably not be any others that are not service specific.  In that respect, CAN ID is service specific and I would like to see that as eeprom(3) with the two NN bytes promoted by one in the eeprom.

src/Configuration.cpp
MartinDaCosta53: Precisely!  As it stands at the moment, the default is no heartbeat which is at variance with the specified position.  It is easily changed if we decide to retain the specified default condition but an opportunity to review.

src/Configuration.h
MartinDaCosta53: If event ack is to be persisted then yes, we should.  I wasn't going to put it in yet until the code is written for the function in configuration.cpp.  Mind you, that could probably be done now, as well!
I think that this flag should be in eeprom(4) as it is a persisted service specific.

src/Configuration.h
MartinDaCosta53: Incidentally, because the Uninitialised / Normal has a value of zero or one (ModuleMode) and not false / true, it is easiest if this occupies the first bit of the byte as this then obviates bit shifting.

src/MinimumNodeService.h
MartinDaCosta53: OK

src/MinimumNodeService.cpp
MartinDaCosta53: This is a default for when we are not MODE_NORMAL.  If we are not Normal than no other services should be doing anything.
Thinking overnight, I think that the "break;" at line 413 should be a default case and not just be hanging.

MartinDaCosta53: I though it was. the else refers to the "if (instantMode == MODE_NORMAL)" statement at line 390

MartinDaCosta53: Sorry, getting confused about which else statement you are referring to.  Are you, in fact, saying that the statement at line 415 is superfluous? I think that is only the case if I get rid of both else statements at lines 415 and 420 and simple end with "return PROCESSED;"  Do you agree?

src/EventTeachingService.cpp
MartinDaCosta53: We don't need to test for Normal mode because if it isn't in Normal mode MNS will have shut it down.

SvenRosvall: Sorry don't understand "promulgation". Assuming you mean that MNS stops other services from handling the MODE code. 
This relies on the "return PROCESSED" in MNS which I commented on elsewhere.

src/MinimumNodeService.cpp
SvenRosvall: OK, I read what you are saying. It does make sense that MODE is only used in Normal mode. 
I am just thinking that in the future things might change. But we'll review the code in that case.

Regarding the "break", I think you can replace it with "return NOT_PROCESSED;". The break relies on such a return statement at the end of the function. It is clearer to have the return NOT_PROCESSED here.

SvenRosvall: No It is just that line 422 should be indented relative to the curly bracket on the line above.

SvenRosvall: I sent you a separate email on the "else" style.

The issue about letting other services see MODE commands in other modes than Normal is something we should discuss separately. And changing this behaviour can wait until we see a need for it.

src/Configuration.cpp
SvenRosvall: Not a problem if you leave eeprom[0] as currentMode and store the flags in eeprom[4].

src/EventTeachingService.cpp
MartinDaCosta53: My understanding was that the services were offered the opcode in sequence.  Since MNS is service 1, it would always be able to do the initial checking of the opcode for length and node number.  If the service, MNS or later, uses the opcode then it will "return PROCESSED" at which point there is no longer a message to offer to subsequent services and the "for" loop exits.

src/MinimumNodeService.cpp
MartinDaCosta53: I would prefer "return NOT_PROCESSED" as it is much clearer as to what is happening.

MartinDaCosta53: Seen the email, which makes a lot of sense and makes things much easier to read.
I agree that we can discuss passing on of the MODE commands.  I also agree that we should wait for a proven need before adding the capability.

src/Configuration.cpp
MartinDaCosta53: I think that we could have a discussion about how we organise the lower eeprom addresses. We do not have to be constrained by the way in which Duncan did it for CBUS.

src/EventTeachingService.cpp
SvenRosvall: That makes sense if MNS is actually the first service in the list. How can we ensure that MNS is actually the first service in the list? Should we tolerate users who don't put MNS first? (I.e. make the system work somewhat, or should we error and stop?)

dpharris:
Can MNS be automatically added to the list in the first position?  Users
would add other Services.

So, from the example,
   VLCB::Controller controller(&userInterface, &modconfig, &can2515,
                            { &canService, &nvService, &ecService,
&epService, &etService });

David

src/EventTeachingService.cpp
MartinDaCosta53: The question here is does "for (Service * service : services)" run through the list in service number order or the sequence that is in the sketch pre-processor listing "VLCB::Controller controller(........)".  If it is the former, then there should be no issue but if it is the latter there is.

Since MNS is the only mandatory service, I have always been concerned about it being called a service at all as it is as fundamental to function as configuration and control. Is it possible to make its inclusion integral to the library so that the user does not have the option to control when it is called or even to inadvertently leave it out?

SvenRosvall: Yes, that for construct will iterate the service list in order.
Yes, I agree. I have been thinking this. I like to have MNS as an object in the user sketch so that it can be configured. Although we don't have any configurations in MNS at this stage. Just don't want to close the door for the future. 
Anyways, let's leave this for a f2f discussion.

src/Configuration.h
SvenRosvall: Yes, It also allows for new major modes to be added in the future. I can't see any at this stage but just don't want to close the door for such a feature. This is why I want to store the heartbeat flag in eeprom[4], together with the eventAck flag.

src/EventTeachingService.cpp
MartinDaCosta53: Agreed for f2f discussion

src/Configuration.h
MartinDaCosta53: I see your thought process and have re-organised accordingly.  However, I would like to see mandated flags, such as heartbeat, kept in a separate location to optional service ones.  I have tried to explain my thoughts is a separate email. We cab discuss at a f2f

PR #37, opened 2023-09-24
Fix memory problem in initializer_list

SvenRosvall:
The memory in initializer_list is a temporary array created by the compiler. This memory needs to stored in a managed container.


MartinDaCosta53:
If I am following this, initialiser_list.h is used for setting up your test regime.  initialiser_list is a temporary file used to organise and assign memory to services during compilation.  Is this correct?

david284 (approved):
Sounds sensible to me

MartinDaCosta53:
Fine with me

PR #39, opened 2023-09-28
Remove friends from Controller class.

SvenRosvall:
Add getter and manipulation methods to Controller class so that service classes don't need to be friends.


MartinDaCosta53 (approved):
I think I follow this.  Once it's merged, I'll look forward to trying it on 4in4out

PR #43, opened 2023-10-05
Minor code simplification, etc

MartinDaCosta53:
Producer service code simplification.
Error correction Event Teaching service
Introduce Example VLCB4in4out.  Documentation to follow but will be based on CANmINnOUT

### DIFF
--- a/docs/VLCB4in4out_README.md
+++ b/docs/VLCB4in4out_README.md
@@ -1,0 +1,180 @@
+# VLCB4IN4OUT
+
+An Arduino program to allocate 4 switches as input pins and 4 outputs to LEDs.
+
+Key Features:
+- MERG VLCB interface.
+- LED flash rate selectable from event variables.
+- Switch function controllable by node variables.
+- Modular construction to ease adaptation to your application.
+
+## Overview
+
+The program is written in C++ but you do not need to understand this to use the program.
+The program includes a library that manages the LED functionality.
+
+## Using VLCB4in4out
+
+The MCP2515 interface requires five Arduino pins to be allocated. Three of these are fixed
+in the architecture of the Arduino processor. One pin must be connected to an interrupt
+capable Arduino pin. The Chip Select pin can be freely defined. This example is configured
+for use with an Uno but can easily be adapted for use with other Arduino types.
+
+**It is the users responsibility that the total current that the Arduino is asked to supply 
+stays within the capacity of the on board regulator.  Failure to do this will result in 
+terminal damage to your Arduino.**
+
+Pins defined as inputs are active low.  That is to say that they are pulled up by an 
+internal resistor. The input switch should connect the pin to 0 Volts.
+
+Pins defined as outputs are active high.  They will source current to (say) an LED. It is 
+important that a suitable current limiting resistor is fitted between the pin and the LED 
+anode.  The LED cathode should be connected to ground.
+
+### Library Dependencies
+
+As well as the VLCB Library suite associated with this example, the following third party
+libraries are required:
+
+Library | Purpose
+---------------|-----------------
+Streaming.h  |*C++ stream style output, v5, (http://arduiniana.org/libraries/streaming/)*
+Bounce2.h    |*Debounce of switch inputs*
+ACAN2515.h   |*library to support the MCP2515/25625 CAN controller IC*
+
+### Application Configuration
+
+The example is configured for use with the MERG Kit 110 CAN Shield. This has a crystal
+frequency of 16MHz.  If you use another CAN2515 board, the crystal frequency may be 8MHz.
+In this case it will be necessary to change the code at or about line 173 appropriately:
+```
+can2515.setOscFreq(16000000UL);   // select the crystal frequency of the CAN module
+```
+
+### VLCB Op Codes
+
+The following Op codes are supported:
+
+OP_CODE | HEX | Function
+----------|---------|---------
+ OPC_ACON | 0x90 | On event
+ OPC_ACOF | 0x91 | Off event
+ OPC_ASON | 0x98 | Short event on
+ OPC_ASOF | 0x99 | Short event off
+
+### Event Variables
+
+#### Consumed Events
+
+Event Variables control the action to take when a consumed event is received.
+The number of Event Variables (EV) is equal to the number of LEDs plus 1.
+
+The remaining Event Variables control the LEDs. 
+Event Variable 2 (EV2) controls the first LED pin in the ```LED``` array. 
+EV3 controls the second LED pin, etc.
+
+The following EV values are defined to control the LEDs:
+
+ EV Value | Function
+--------|-----------
+ 0 | LED off
+ 1 | LED on
+ 2 | LED flash at 500mS
+ 3 | LED flash at 250mS
+
+#### Produced Events
+
+Event Variables, whilst present, are not used in this application.  If used in an 
+application, they may be set in the normal manner using the FCU.
+
+### Node Variables
+
+Node Variables control the action to take when a switch is pressed or depressed.
+
+The number of Node Variables (NV) is equal to the number of switches.
+NV1 corresponds to the first pin defined in the array ```SWITCH```, 
+NV2 corresponds to the second pin in that array, etc.
+
+The following NV values define input switch function:
+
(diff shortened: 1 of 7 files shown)


examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
```diff
@@ -0,0 +1,550 @@
...
+
+    No additional restrictions : You may not apply legal terms or technological measures that
+                                 legally restrict others from doing anything the license permits.
+
+   ** For commercial use, please contact the original copyright holder(s) to agree licensing terms
+
+    This software is distributed in the hope that it will be useful, but WITHOUT ANY
+    WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE
+
+*/
```
SvenRosvall, line 40: I prefer the shorter copyright notice we have in the other examples.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
```diff
@@ -0,0 +1,550 @@
...
+}
+
+void processSwitches(void) {
+  bool isSuccess = true;
+  for (byte i = 0; i < NUM_SWITCHES; i++) {
+    moduleSwitch[i].update();
+    if (moduleSwitch[i].changed()) {      
+      byte nn_en[4];
+      modconfig.readEvent(i, nn_en);
+      byte evval = modconfig.getEventEVval(i, 1);
```
SvenRosvall, line 267: Hmm, so here you use EV1 in the produced event. 
I think that in CANmINnOUT we used node variables for the switch behaviours, right?
Is it possible to set EV1 in these produced events from FCU?

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
```diff
@@ -0,0 +1,550 @@
...
+
+          break;
+
+        default:
+          DEBUG_PRINT(F("> Invalid EV value."));
+          break;
+      }
+    }
+  }
+  if (!isSuccess) {
```
SvenRosvall, line 318: Always false. isSuccess is never set to false.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
```diff
@@ -0,0 +1,550 @@
+
+// CANmINnOUT
```
SvenRosvall, line 2: A left-over from previous code?

examples/VLCB4IN4OUT/LEDControl.h
SvenRosvall, line 1: I would like a comment section that describes how this is used and what event variables and node variables do.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
MartinDaCosta53: Yes, so do I.  Never thought to look at the copyright notice!  I'll change it.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
MartinDaCosta53: Yes.  It is as easy to change EV Values for produced events as it is consumed events in the FCU.  Just double click on them at the default window comes up.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
MartinDaCosta53: Fair point.  I'll delete it.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
MartinDaCosta53: Yes!  Thanks.

examples/VLCB4IN4OUT/LEDControl.h
SvenRosvall: OK, that will be good.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
MartinDaCosta53: Produced events are now set as NVs and accessed through the application sketch.  This is to make life easier for COE (I think!)

examples/VLCB4IN4OUT/LEDControl.h
MartinDaCosta53: Now done

docs/VLCB4in4out_README.md
SvenRosvall, line 1: Thinking that this should be located in the examples directory beside the code. Not sure though. What do you think?

MartinDaCosta53: We could do.  I put it in docs so that all explanation material was in one place, but I have no strong feelings about it.  I can see a logic to moving it.

examples/VLCB_1in1out/VLCB_1in1out.ino
```diff
@@ -205,7 +205,7 @@ void processModuleSwitchChange()
   if (moduleSwitch.stateChanged())
   {
     bool state = moduleSwitch.isPressed();
-    byte eventNumber = 1;
+    byte eventNumber = 0;  // First switch has an event index of 0
```
SvenRosvall, line 208: Are you sure about this? 
It doesn't really matter as this example sketch can do what it wants. 
Just curious why you decided to change?
It looks a bit strange to have an EN == 0.

src/EventTeachingService.cpp
```diff
@@ -448,11 +448,15 @@ Processed EventTeachingService::handleLearnEventIndex(CANFrame *msg)
...
         module_config->readEvent(index, eventTableNNEN);
-        if (memcmp(eventTableNNEN, &msg->data[1], 4) == 0)
+        if (memcmp(eventTableNNEN, &msg->data[1], 4) != 0)
         {
           module_config->writeEvent(index, &msg->data[1]);
         }
 
+        if (evIndex == 0)  // Not a valid evIndex
+        {
+          evIndex = 1;
```
SvenRosvall, line 458: I think I would respond with an error message instead of changing the input data. It might be confusing when a different event variable is changed than what the user exected.

docs/VLCB4in4out_README.md
SvenRosvall: We can look at this at a later stage. 
I also want to review how we link to the example sketches from the README.md file.

docs/VLCB4in4out_README.md
```diff
@@ -0,0 +1,183 @@
...
+
+The program is written in C++ but you do not need to understand this to use the program.
+The program includes a library that manages the LED functionality.
+
+## Using VLCB4in4out
+
+The MCP2515 interface requires five Arduino pins to be allocated. Three of these are fixed
+in the architecture of the Arduino processor. One pin must be connected to an interrupt
+capable Arduino pin. The Chip Select pin can be freely defined. This example is configured
+for use with an Uno but can easily be adapted for use with other Arduino types.
```
SvenRosvall, line 21: I think it would be appropriate to describe what the fixed pins are and what pins are used in the sketch for the ones you can change. Also for pins like INT you can only choose pin 2 or 3.
You may also make a reference to to your shield where certain pins are used.

docs/VLCB4in4out_README.md
SvenRosvall, line 1: Lots of good stuff here that will be useful for the other example sketches too.

examples/VLCB4IN4OUT/VLCB4IN4OUT.ino
```diff
@@ -0,0 +1,519 @@
...
+///  setup VLCB - runs once at power on called from setup()
+//
+void setupVLCB()
+{
+  // set config layout parameters
+  modconfig.EE_NVS_START = 10;
+  modconfig.EE_NUM_NVS = NUM_SWITCHES;
+  modconfig.EE_EVENTS_START = 50;
+  modconfig.EE_MAX_EVENTS = 64;
+  modconfig.EE_PRODUCED_EVENTS = 0;
```
SvenRosvall, line 133: You have 4 switches that may produce events. Shouldn't EE_PRODUCED_EVENTS be 4 here?

PR #44, opened 2023-10-10
Improved method of reporting NAME in Teach Mode

MartinDaCosta53:


src/EventTeachingService.cpp
```diff
@@ -91,26 +91,6 @@ Processed EventTeachingService::handleMessage(unsigned int opc, CANFrame *msg)
...
-      msg->data[0] = OPC_NAME;
-      memcpy(msg->data + 1, controller->getModuleName(), 7);
-      controller->sendMessage(msg);
-    }
-    return PROCESSED;
-    
-  default:
-    // unknown or unhandled OPC
-    // DEBUG_SERIAL << F("> opcode 0x") << _HEX(opc) << F(" is not currently implemented")  << endl;
-    return NOT_PROCESSED;
```
SvenRosvall, line 113: Hmm, I think you should leave in this default bit.
You should get a compile error as there isn't a return in all branches.

MartinDaCosta53: But it has gone back to the way it was before.  EventTeachingService no longer needs to see OPC_RQMN.

SvenRosvall: Sure, it is the default: part that needs to be kept. I.e. keep lines 110-113.
It shouldn't compile without those lines.

MartinDaCosta53: Oh, right. I see what you mean.  Now restored.

PR #45, opened 2023-10-14
Updates to MODE command

SvenRosvall:
Changes after adding unit tests for handling the MODE op-code.
These changes match my understanding of the current MODE as described in the MODE proposal document.

### DIFF
--- a/src/Configuration.h
+++ b/src/Configuration.h
@@ -38,6 +38,10 @@ enum ModuleMode {
   MODE_UNINITIALISED = 0,
   MODE_NORMAL = 1,
   MODE_SETUP = 2,
+  
+  // Minor modes
+  MODE_HEARTBEAT_ON = 0x0C,
+  MODE_HEARTBEAT_OFF = 0x0D,
 };
 
 //
--- a/src/Configuration.h
+++ b/src/Configuration.h
@@ -35,9 +35,9 @@ enum FlagBits {
 /// Controller modes
 //
 enum ModuleMode {
-  MODE_UNINITIALISED = 0,
+  MODE_UNINITIALISED = 0xFF,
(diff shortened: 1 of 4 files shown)


MartinDaCosta53 (commented):
I agree with changing the cases to use constants, but where are these defined?  I can only see 3 defined in configuration.h in the enum ModuleMode.  Should the complete list be in vlcbdefs.hpp?

MartinDaCosta53 (commented):
Oh, OK. Found them in the configuration.h changes.  We need to add Learn On and Learn Off

MartinDaCosta53 (commented):
Oh, OK. Found them in the configuration.h changes.  We need to add Learn On and Learn Off

MartinDaCosta53 (commented):
Save for making the enum ModuleMode in configuration.h fully comprehensive, this all looks fine to me.

SvenRosvall:
> I agree with changing the cases to use constants, but where are these defined? I can only see 3 defined in configuration.h in the enum ModuleMode. Should the complete list be in vlcbdefs.hpp?

Hmm, I added MODE_HEARTBEAT_ON/OFF to Configuration.h in this change. Is it not visible?
Yes, I think this list should be added to vldbdefs.hpp. I'll do that when the MODE proposal is accepted.

SvenRosvall:
> Save for making the enum ModuleMode in configuration.h fully comprehensive, this all looks fine to me.

I only added the MODE parameters that were needed in MNS. When they are added to VLCBDEFS I'll add the whole list.

MartinDaCosta53:
EventTeachingService uses Learn Mode on and off and I think that we should update that from using hex numbers at the same time.  Why not just put the whole lot in vlcbdefs now?

PR #46, opened 2023-10-15
Move module MODE constants to vlcbdefs.hpp.

SvenRosvall:


src/vlcbdefs.hpp
MartinDaCosta53, line 1: Why can't we call the enum VlcbModeParams?  I don't like to see us perpetuating the use of Cbus unless absolutely necessary for compatability.  I don't think that this is the case, here.

SvenRosvall: I was looking at the other types which still have the "Cbus" prefix. I have kept these as it makes it easy to incorporate any changes that is done to CBUS-defs.

But now I see that we have used "Vlcb" for things that are specific for VLCB. Had forgotten this. Will fix this.

SvenRosvall: Changed to VlcbModeParams

PR #47, opened 2023-10-18
Simplify SD indexing.

SvenRosvall:


MartinDaCosta53 (approved):
That is a much neater way of finding a specific service rather than searching

PR #48, opened 2023-10-18
Code trimming

SvenRosvall:
Some small changes to reduce code size both in source code and compiled code.

### DIFF
--- a/src/CAN2515.cpp
+++ b/src/CAN2515.cpp
@@ -19,17 +19,12 @@ namespace VLCB
 /// constructor
 //
 CAN2515::CAN2515()
+  : _num_rx_buffers(NUM_RX_BUFFS)
+  , _num_tx_buffers(NUM_TX_BUFFS)
+  , _csPin(MCP2515_CS)
+  , _intPin(MCP2515_INT)
+  , _osc_freq(OSCFREQ)
 {
-  initMembers();
-}
-
-void CAN2515::initMembers()
-{
-  _num_rx_buffers = NUM_RX_BUFFS;
-  _num_tx_buffers = NUM_TX_BUFFS;
-  _csPin = MCP2515_CS;
-  _intPin = MCP2515_INT;
-  _osc_freq = OSCFREQ;
 }
 
 //
--- a/src/CAN2515.h
+++ b/src/CAN2515.h
@@ -66,7 +66,6 @@ public:
   unsigned int _numMsgsSent, _numMsgsRcvd;
 
 private:
-  void initMembers();
   Controller * controller;
   unsigned long _osc_freq;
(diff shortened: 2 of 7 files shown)


MartinDaCosta53 (commented):
I'm wondering if line 72 is required.  The only time heartbeat is looked at is by MNS in the begin() function.  So, setting heartbeat = beat is superfluous as it is never used.

MartinDaCosta53 (commented):
Where do I find stdarg.h?  Does this subsume vlcbdefs.hpp?

SvenRosvall:
> Where do I find stdarg.h? Does this subsume vlcbdefs.hpp?

<stdarg.h> is a standard header. It is included in the Arduino IDE. It is used for the function sendMessageWithNNandData() on line 296 in Controller.h where we have the "..." for "any number of parameters". the names va_xxx are defined in stdarg.h and are used to handle those extra parameters. Let me know if you want an explanation how that works.

I took out vldbdefs.hpp here as it is already included by other header files.

These changes are not related.

SvenRosvall:
> I'm wondering if line 72 is required in Configuration.ccp. The only time heartbeat is looked at is by MNS in the begin() function. So, setting heartbeat = beat is superfluous as it is never used. Same goes for line 87 eventAck = ea

Yes, looks like that can be reworked. There is a noHeartbeat in MNS. I guess the begin() function can look at the flag bit in module_config. Will look into this. But need your advice as you are the one who created this code.

MartinDaCosta53:
I'll take it out and test it.  I'm pretty convinced that they're superfluous

MartinDaCosta53:
> > I'm wondering if line 72 is required in Configuration.ccp. The only time heartbeat is looked at is by MNS in the begin() function. So, setting heartbeat = beat is superfluous as it is never used. Same goes for line 87 eventAck = ea
> 
> Yes, looks like that can be reworked. There is a noHeartbeat in MNS. I guess the begin() function can look at the flag bit in module_config. Will look into this. But need your advice as you are the one who created this code.

I'll take it out and test it. I'm pretty convinced that they're superfluous

MartinDaCosta53:
> > Where do I find stdarg.h? Does this subsume vlcbdefs.hpp?
> 
> <stdarg.h> is a standard header. It is included in the Arduino IDE. It is used for the function sendMessageWithNNandData() on line 296 in Controller.h where we have the "..." for "any number of parameters". the names va_xxx are defined in stdarg.h and are used to handle those extra parameters. Let me know if you want an explanation how that works.
> 
> I took out vldbdefs.hpp here as it is already included by other header files.
> 
> These changes are not related.

That's rather clever.  I'd not heard of this before.

PR #52, opened 2023-10-26
Introduce Consume Own Events

MartinDaCosta53:


src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (size > 0)
+  {
+    return true;
+  }
+  else
+  {
+  return false;
+  }
+
+ // return (size > 0);
```
SvenRosvall, line 32: Why did you do the complex if-statement that does exactly the same as the commented out return statement below?

MartinDaCosta53: I spelling things out whilst I tried to get the compiled code to actually work!  I forgot to change this back.

src/ConsumeOwnEventsService.h
```diff
@@ -0,0 +1,61 @@
...
+  unsigned int gets();
+  unsigned int overflows(void);
+    
+private:
+  Controller *controller;
+  Configuration *module_config;  // Shortcut to reduce indirection code.
+  
+  byte numItems = 4;
+  
+  VlcbMessage *buffer = malloc(numItems * sizeof(VlcbMessage));
```
SvenRosvall, line 47: This is a non-trivial expression. The malloc call should be in a constructor. And I would prefer to use "new" instead of the malloc call. "new" is C++ and malloc() is C. ;) 
Also it doesn't compile outside the Arduino IDE as the return value from malloc is "void *" and must be cast to VlcbMessage*.
Leave this one with me.

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/// store an item to the buffer - overwrite oldest item if buffer is full
+/// never called from an interrupt context so we don't need to worry about interrupts
+
+void ConsumeOwnEventsService::put(const VlcbMessage * msg)
+{
+//  VlcbMessage msg;
```
SvenRosvall, line 40: You need to check if the buffer is full before you put something in there.
What should happen when the buffer overflows? Do we overwrite the oldest or do we ignore this last one?

src/ConsumeOwnEventsService.h
```diff
@@ -0,0 +1,61 @@
...
+  
+  byte head = 0;
+  byte tail = 0;
+  byte hwm = 0;
```
SvenRosvall, line 51: "hwm" - cryptic name. What does it do? Can we find a better name?

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+  memcpy((VlcbMessage*)&buffer[head], (const VlcbMessage *)msg, sizeof(VlcbMessage));
```
SvenRosvall, line 41: Why do you need the casts? (VlcbMessage*) and (const VLCBMessage *) These cast to the same type.

src/ConsumeOwnEventsService.h
SvenRosvall, line 47: This non-trivial expression used in this assignment should be in the constructor. 
Also, prefer to use "new" instead of "malloc()". malloc() is C. ;)
Leave this one with me.

src/ConsumeOwnEventsService.h
```diff
@@ -0,0 +1,61 @@
...
+  byte capacity = numItems;
```
SvenRosvall, line 52: Do you need numItems? It is only used here and in the malloc() call above. You can use capacity in malloc() instead.

src/ConsumeOwnEventsService.h
MartinDaCosta53: OK. Thanks

src/ConsumeOwnEventsService.h
```diff
@@ -0,0 +1,61 @@
...
+  byte bufUse(void);
+  unsigned int puts();
+  unsigned int gets();
+  unsigned int overflows(void);
+    
+private:
+  Controller *controller;
+  Configuration *module_config;  // Shortcut to reduce indirection code.
+  
+  byte numItems = 4;
```
SvenRosvall, line 45: Should we allow users to change the size of the buffer? Add a parameter to the constructor?

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+  // if the buffer is full, this put will overwrite the oldest item
+
+  if (full)
+  {
+    tail = (tail + 1) % capacity;
+    ++numOverflows;
+  }
+
+  head = (head + 1) % capacity;
+  full = head == tail;
```
SvenRosvall, line 43: "full" becomes true when the buffer is empty.

src/ConsumeOwnEventsService.cpp
MartinDaCosta53: It's a circular buffer and so if it is full the oldest gets overwritten.  Hopefully, at the rate of production of events and with a buffer size of 4, it will never fill up.

src/ConsumeOwnEventsService.h
MartinDaCosta53: This is head of circular buffer and tail of circular buffer.  HWM is the High Water Mark and it's never reported to anything.  I meant to take it out as superfluous.

src/ConsumeOwnEventsService.cpp
MartinDaCosta53: That's what comes of copying things and then getting overwhelmed with joy when it finally works!  I should gone back to look for such things and taken them out.  Now changed

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+  size = 0;
+
+  return;
+}
+
+/// recalculate number of items in the buffer
+
+byte ConsumeOwnEventsService::bufUse(void)
+{
+  size = capacity;
```
SvenRosvall, line 100: Why are you assigning the object field here, and then assign the return value to "size". I would return the expression you set "size" to here. And if(full) return capacity.
Or rename the function to updateSize() and don't return anything.

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+      size = capacity + head - tail;
+    }
+  }
+
+  return size;
+}
+
+/// number of puts
+
+unsigned int ConsumeOwnEventsService::puts(void)
```
SvenRosvall, line 118: If this is for diagnostics, I would like to name this function getNumberOfPuts().
"puts" is very similar to "put".

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+  return numPuts;
+}
+
+/// number of gets
+
+unsigned int ConsumeOwnEventsService::gets(void)
```
SvenRosvall, line 125: rename this too.

src/ConsumeOwnEventsService.cpp
SvenRosvall: OK, I read the comment now.

src/Controller.h
```diff
@@ -26,7 +26,7 @@ namespace VLCB
 //
 struct VlcbMessage
 {
-  uint8_t len; // Value 0-7 or -1 for messages handled in CanTransport
+  uint8_t len; // Value 0-7 or FF for messages handled in CanTransport
```
SvenRosvall, line 29: Ouch, did I forget to update the comment? Thanks for fixing it.

src/EventConsumerService.cpp
```diff
@@ -51,11 +51,29 @@ void EventConsumerService::processAccessoryEvent(VlcbMessage *msg, unsigned int
   }
 }
 
+void EventConsumerService::process(UserInterface::RequestedAction requestedAction)
+{
+  if (coeService && (coeService->available()))
+  {
+    // DEBUG_SERIAL << ">Getting COE Message " << endl;
+    VlcbMessage *msg = coeService->get();
```
SvenRosvall, line 59: First I thought the messages from the COE Service buffer would go to Controller for normal processing of messages. But you are right here in that we are only interested in event messages.
Just a bit confusing that the process() call is retrieving messages and handling them where we have handleMessage for that.

src/ConsumeOwnEventsService.h
MartinDaCosta53: I thought about this but decided to get some experience on how the buffer behaves.  To that end, I will add a monitor for puts, gets and overflows.  So far, I've only seen the size transition though 1 and back to zero as the buffer gets emptied almost as soon as it has something in it.  If the loop starts to take a long time, it could be a problem.

src/EventConsumerService.h
```diff
@@ -1,11 +1,12 @@
-// Copyright (C) Sven Rosvall (sven@rosvall.ie)
+// Copyright (C) Martin Da Costa 2023 (martindc.merg@gmail.com)
 // This file is part of VLCB-Arduino project on https://github.com/SvenRosvall/VLCB-Arduino
```
SvenRosvall, line 2: Sorry about that. I think I just copied the copyright notice when I shrunk the copyright notice you had here first. Thanks for fixing.

src/ConsumeOwnEventsService.cpp
```diff
@@ -0,0 +1,147 @@
...
+  if (full)
+  {
+    tail = (tail + 1) % capacity;
+    ++numOverflows;
+  }
+
+  head = (head + 1) % capacity;
+  full = head == tail;
+  size = bufUse();
+  hwm = (size > hwm) ? size : hwm;
```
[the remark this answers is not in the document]
MartinDaCosta53: Not a lot,  I need to take it out as it is superfluous.  It was available in a "get" function but never read.

src/ConsumeOwnEventsService.h
SvenRosvall: Consider something CANCOMPUTE or CANGATE where you could get many output events when an input changes.

src/ConsumeOwnEventsService.h
SvenRosvall: Ah. High water mark is a useful metric to know if the buffer need to be resized. 
Yes, the overflow count would tell you that you must resize the buffer. But what if you are close to capacity. This is where high watermark is useful. Please keep it for diagnostics.

src/EventConsumerService.cpp
```diff
@@ -51,11 +51,29 @@ void EventConsumerService::processAccessoryEvent(VlcbMessage *msg, unsigned int
...
+    // DEBUG_SERIAL << ">COE Message handled " << endl;
+  }
+}
+
 Processed EventConsumerService::handleMessage(unsigned int opc, VlcbMessage *msg) 
 {
+  return handleEventMessage(msg);
+}
+
+Processed EventConsumerService::handleEventMessage(const VlcbMessage *msg)
```
SvenRosvall, line 70: The "const" causes problems here. You cannot call processAccessoryEvent() with a const pointer as it needs a non-const pointer. Simply remove "const" in this parameter list.

src/ConsumeOwnEventsService.cpp
MartinDaCosta53: Actually, this whole thing is convoluted in a most unnecessary way. If the buffer is full, it equals capacity, which is numItems.  If it's not full, then that is the only time we need to do anything.  Let me think about this one.

src/ConsumeOwnEventsService.cpp
MartinDaCosta53: Good idea.  I'll change it.

src/ConsumeOwnEventsService.cpp
MartinDaCosta53: Of course.

src/ConsumeOwnEventsService.h
MartinDaCosta53: I agree.  Changed it

src/EventConsumerService.cpp
```diff
@@ -51,11 +51,29 @@ void EventConsumerService::processAccessoryEvent(VlcbMessage *msg, unsigned int
...
+    // DEBUG_SERIAL << ">Getting COE Message " << endl;
+    VlcbMessage *msg = coeService->get();
+    bool done = handleEventMessage(msg);
+    // DEBUG_SERIAL << ">COE Message handled " << endl;
+  }
+}
+
 Processed EventConsumerService::handleMessage(unsigned int opc, VlcbMessage *msg) 
 {
+  return handleEventMessage(msg);
```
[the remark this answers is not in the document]
MartinDaCosta53: handleEventMessage is also called from process().  This doesn't extract opc from the get().   That's why I did it like this.

src/EventConsumerService.cpp
MartinDaCosta53: Yes, I originally sent it to Controller but then decided that it was just creating a complex route.

src/EventConsumerService.cpp
MartinDaCosta53: OK

src/ConsumeOwnEventsService.h
MartinDaCosta53: Fair point.  I'll leave the default at 4 and add an option to specify something else.

src/EventConsumerService.h
MartinDaCosta53: No worries

src/Controller.h
MartinDaCosta53: No worries.

src/ConsumeOwnEventsService.cpp
SvenRosvall: I am ok with the logic. What I am commenting on is that you both set the field "size" and return a value which is then used to assign to "size". Duplicate work.

MartinDaCosta53: Yes, I agree that it does duplicate and become overcomplicated.  I'll have a look at it later this morning in hospital before the chemicals get to me!

PR #56, opened 2023-11-02
Added SerialGC

david284:
Please review this code


src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+        }
+        // mark as extended message
+        strcpy (gcBuffer,":X");
+        // extended 29 bit CAN idenfier in bytes 2 to 9
+        // chars 2 & 3 are ID bits 21 to 28
+        sprintf(gcBuffer + 2, "%02X", (msg->id) >> 21);
+        // char 4 -  bits 1 to 3 are ID bits 18 to 20
+        sprintf(gcBuffer + 4, "%01X", ((msg->id) >> 17) & 0xE);
+        // char 5 -  bits 0 to 1 are ID bits 16 & 17
+        sprintf(gcBuffer + 5, "%01X", ((msg->id) >> 16) & 0x3);
```
SvenRosvall, line 73: Are these bits correct? Looks like the 29 bits are not contiguous in the hex string.

src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+        offset = 10;
+      } else {// mark sas standard message
+        if (msg->id > 0x7FF)
+        {
+          // id is greater than 11 bits, so fail the encoding
+          return false;
+        }
+        strcpy (gcBuffer,":S");
+        // standard 11 bit CAN idenfier in bytes 2 to 5, left shifted 5 to occupy highest bits
+        sprintf(gcBuffer + 2, "%04X", msg->id << 5);
```
SvenRosvall, line 85: You could combine the strcpy() and the sprintf() by using a format string like ":S%04X".

src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+        offset = 6;
+      }
+      // set RTR or normal - byte 6 or 10
+      if (msg->rtr) {
+        strcpy (gcBuffer + offset++,"R");
+      } else {
+        strcpy (gcBuffer + offset++,"N");
+      }
+      // add terminator in case len = 0, will be overwritten if len >0
+      strcpy (gcBuffer + offset,";");
```
SvenRosvall, line 95: I would add the ";" to the buffer after the loop. That means the ";" is added regardless of the data length.

src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+        return false;
+      }
+      // ok, all hex, so build up id from characters 2 to 9
+      // chars 2 & 3 are bits 21 to 28
+      message->id = uint32_t(ascii_pair_to_byte(&gcBuffer[2])) << 21;
+      // chars 4 & 5 -  bits 5 to 7 are bits 18 to 20
+      message->id += uint32_t(ascii_pair_to_byte(&gcBuffer[4]) & 0xE0) << 13;
+      // chars 4 & 5 -  bits 0 to 1 are bits 16 & 17
+      message->id += uint32_t(ascii_pair_to_byte(&gcBuffer[4]) & 0x3) << 16;
+      // chars 6 & 7 are bits 8 to 18
```
SvenRosvall, line 168: Typo:
... 8 to 16

src/Parameters.h
```diff
@@ -15,7 +15,7 @@ class Parameters
   explicit Parameters(Configuration const & config)
   {
     params[PAR_NUM] = 20;                     //  0 num params = 20
-    params[PAR_MANU] = MANU_VLCB;              //  1 manf = MERG, 165
+    params[PAR_MANU] = MANU_MERG;              //  1 manf = MERG, 165
```
SvenRosvall, line 18: This is for testing. Please don't commit.

src/SerialGC.cpp
```diff
@@ -0,0 +1,164 @@
...
+// 
+// For an extended can message with an 29 bit identifier, the GridConnect identifier field is 8 hex characters,
+// i.e. 32 bits
+// In this case the mapping is a bit more complex
+// IDENTIFIER[0,1] - bits [28:21] of the can identifier
+// IDENTIFIER[2] - bits [20:18] of the can identifier, left shifted by 1, bottom bit set to zero
+// IDENTIFIER[3] - bits [17:16] of the can identifier, bottom two bits, top two bits set to 0
+// IDENTIFIER[4,5,6,7] - bits [15:0] of the can identifier
+// (This maps directly onto SIDH, SIDL, EIDH and EIDL registers the PIC processor family)
+// And the GridConnect Identifier field is also leading zero padded, so always 8 characters for an extended message
```
SvenRosvall, line 53: This comment block with explanation of the GC format is already in GridConnect.cpp. Don't need it twice.

src/SerialGC.cpp
```diff
@@ -0,0 +1,164 @@
...
+  // return true if valid message assembled & ready
+  //
+  bool SerialGC::available()
+  {
+    bool result = false;
+    static int rxIndex = 0;
+    if (Serial.available())
+    {     
+      char c = Serial.read();
+      if(c >= 'a' && c <= 'z') bitClear(c,5);   // ensure letters are upper case
```
SvenRosvall, line 81: You can use "c = tolower(c)". Then you don't need the comment.

src/SerialGC.h
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) David Ellis
+// This file is part of VLCB-Arduino project on https://github.com/SvenRosvall/VLCB-Arduino
+// Licensed under the Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License.
+// The full licence can be found at: http://creativecommons.org/licenses/by-nc-sa/4.0/
+
+#pragma once
+
+// header files
+#include <Arduino.h>
+#include <Controller.h>
```
SvenRosvall, line 10: Do you need to include <Controller.h> ?

src/GridConnect.cpp
david284: Thats correct - thats the difference between the 'official' GridConnect and the 'CBUS' variant of Grid connect
The hex values are mangled to fit the can controller used in PIC's, so for 11 bit ID's the lower 5 bits aren't used, and for 29 bit ID's there are three unused bits in the 3rd byte
I've tried to show this in the notes at the start of the file - but I've found it hard to describe how it's split
That's why the unit tests where so important, to be able to test the boundaries of this split - I think the unit tests will show this split better than words

src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+        // standard 11 bit CAN idenfier in bytes 2 to 5, left shifted 5 to occupy highest bits
+        sprintf(gcBuffer + 2, "%04X", msg->id << 5);
+        offset = 6;
+      }
+      // set RTR or normal - byte 6 or 10
+      if (msg->rtr) {
+        strcpy (gcBuffer + offset++,"R");
+      } else {
+        strcpy (gcBuffer + offset++,"N");
+      }
```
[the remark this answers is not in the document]
david284: Thats neater - I've changed that

test/testGridConnect.cpp
```diff
@@ -0,0 +1,222 @@
...
+  testGridConnectDecode_DATA(":X00000000N01;", 1, true);                    // extended msg, 1 data byte
+  testGridConnectDecode_DATA(":X00000000N0102030405060708;", 8, true);      // extended msg, 8 data bytes
+  testGridConnectDecode_DATA(":X00000000N0;", 1, false);                    // extended msg, wrong number of data chars
+  testGridConnectDecode_DATA(":X00000000N010203040506070;", 8, false);      // extended msg, wrong number of data chars
+  testGridConnectDecode_DATA(":X00000000N010203040506070809;", 9, false);   // extended msg, too many data bytes
+  testGridConnectDecode_DATA(":X00000000NQ102030405060708;", 8, false);     // extended msg, invalid char in data
+  testGridConnectDecode_DATA(":X00000000N010203040506070Q;", 8, false);     // extended msg, invalid char in data
+
+
+}
```
SvenRosvall, line 222: Please also add some tests with lower case characters, both hex and the S/X and N/R bits.

src/GridConnect.cpp
david284: Yep, thats neater - done that

src/Parameters.h
david284: Reverted to previous version of file

src/SerialGC.cpp
david284: Yep, needs removing - done that

test/testGridConnect.cpp
```diff
@@ -0,0 +1,222 @@
...
+  // test encoding extended ID - 29 bits, max Id 0x1FFFFFFF
+  // encoded with bits 18 to 28 shifted by 3
+  // so need to test around this gap
+  testGridConnectEncode_ExtendedID(0x0, ":X00000000N0102;", true);        // all bits zero
+  testGridConnectEncode_ExtendedID(0x1, ":X00000001N0102;", true);        // just bit 0 set
+  testGridConnectEncode_ExtendedID(0x20000, ":X00020000N0102;", true);    // bit 17 - before break in coding
+  testGridConnectEncode_ExtendedID(0x3FFFF, ":X0003FFFFN0102;", true);    // all up to bit 17 - 18 bits
+  testGridConnectEncode_ExtendedID(0x40000, ":X00200000N0102;", true);    // bit 18 - after break in coding
+  testGridConnectEncode_ExtendedID(0x1FFC0000, ":XFFE00000N0102;", true); // bits 18 to 28 - 11 bits
+  testGridConnectEncode_ExtendedID(0x1FFFFFFF, ":XFFE3FFFFN0102;", true); // all 29 bits set
```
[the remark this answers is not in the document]
david284: No problem, any feedback is useful

src/SerialGC.h
david284: I copied the includes from CAN2515.h, so didn't check that
However, if I remove it,. the build fails
I suspect CanTransport requires it

src/GridConnect.cpp
david284: yep, I do like keeping these two separate as they are effectively two different 'fields' in the hex string - but happy to change it if you think it's better

src/GridConnect.cpp
david284: well spotted...

src/GridConnect.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Do data segment - convert hex array to byte array
+    // find out how many chars in data segment (0 to 16, in multiples of 2) 
+    // should be gcBufferLength minus gcIndex as it is now (after RTR flag), minus 1
+    int dataLength = gcBufferLength - gcIndex - 1;
+    // must be even number of hex characters, and no more than 16
+    if ((dataLength % 2 ) || (dataLength > 16 )) 
+    {
+      return false;
+    } 
+    else
```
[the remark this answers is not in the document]
david284: yep - hangover from previous style - changed

src/SerialGC.cpp
david284: For some reason, VSCode doesn't recognise 'lotower' (probably another library issue), but as I no longer need to compile SerialGC to run the GridConnect tests, I can build & run the tests & it doesn't cause a problem, and it does compile fine in Ardiuno IDE, so changed that

david284: Sorry - meant to say 'toupper' as thats what I wanted - but same comment applies

test/testGridConnect.cpp
david284: Done that - now GridConnect Decodes will fail if lower case present (as per GridConnect format)

src/GridConnect.cpp
SvenRosvall: Yes, the unit test explains it all.

src/SerialGC.cpp
SvenRosvall: toupper() should be defined in <ctype.h>, which you have included. So this should work. Strange.

src/SerialGC.h
SvenRosvall: Yes, CanTransport.h needs it. It should include <Controller.h>. Then it can be removed from here.
Same effect but neater.

test/testGridConnect.cpp
SvenRosvall: Hmm, I thought they would succeed. but now I see that the toupper() call is in SerialGC instead.

PR #57, opened 2023-11-04
Implement a Serial UI

SvenRosvall:
Move the serial processing to a dedicated UserInterface implementation that can be chosen by the user.


fletchjp:
I have been waiting for this. Is there an example or should I start one?

MartinDaCosta53:
This is a good addition to the library range that removes code that many users will not understand from their sketches and ensuring that library links are maintained.  However, I am concerned that this is tied to CAN2515 and is thus not transport agnostic.  I assume that this is to allow for switch cases 'c' and 'y'.  Case 'c' doesn't do anything as the code has been commented out in CAN2515.  It could, therefore, be removed.  Case 'y' should be redirected to a transport reset.  This would ensure that we do not have a problem going forward.

PR #59, opened 2023-11-08
Remove FrameHandler.

SvenRosvall:
Framehandler is only used by some example sketches to print out incoming messages.


MartinDaCosta53:
This should all be included in an appropriate service and so I agree that it needs to come out of here.

MartinDaCosta53:
Approved

PR #60, opened 2023-11-08
Major change to Event Producer to use EV1

MartinDaCosta53:
Following discussion on 5 Nov 2023 in the VLCB Team, the method of generating and using produced events has been changed.  Event Variable 1 now has a value of zero for a consumed event or the index of a producer activity for a produced event.  The system is that used by Ian Hogg in his Universal software.
This change results in a much simpler interface with the FCU that allows drag and drop teaching techniques to be used.

### DIFF
--- a/docs/VLCB4in4out_README.md
+++ b/docs/VLCB4in4out_README.md
@@ -82,13 +82,16 @@ OP_CODE | HEX | Function
 
 ### Event Variables
 
+Event Variable 1 (EV1) is reserved for system use to identify the source of
+produced event.
+
 #### Consumed Events
 
 Event Variables control the action to take when a consumed event is received.
-The number of Event Variables (EV) is equal to the number of LEDs.
+The number of Event Variables (EV) is equal to the number of LEDs plus one.
 
-Event Variable 1 (EV1) controls the first LED pin in the ```LED``` array. 
-EV2 controls the second LED pin, etc.
+Event Variable 2 (EV2) controls the first LED pin in the ```LED``` array. 
+EV3 controls the second LED pin, etc.
 
 The LEDs are controlled by the LEDControl class.  This allows for any LED to be
 switched on, switched off or flashed at a rate determined in the call:
@@ -105,8 +108,22 @@ The following EV values are defined to control the LEDs in this example:
 
 #### Produced Events
 
-Event Variables, whilst present, are not used in this application.  If used in an 
-application, they may be set in the normal manner using the FCU.
+The Events Table in an Uninitiailised module is empty. When the module changes to
+Normal Mode from Uninitialised, the Events Table is populated with a default
+producer event for each of the four switches (or, in general, each producer item).
+These default events can be removed by using OPC_EVULN whist the module is in
+learn mode.
+
+A module can also be taught a short event or, indeed, a spoof event. This is
+easily done using any of the normal FCU teach techniques.  For example, a short
+event can be dragged from the Software Node and dropped onto the VLCB4in4out in
+the Node Window.  When the event variable dialgue opens, put the index number of
+the switch to be associated with the event in EV1 and press OK.  The selected
+switch will now generate that short event.
+
+If a switch default has been unlearnt and that switch not been assigned to an
+event by the FCU, operation of the switch will result in a new default event
+being generated.
 
 #### Consume Own Events
 
@@ -114,9 +131,10 @@ If the Produced Events Service and the Consume Events Service are both applied,
 the Consume Own Events Service can also be enabled.  This service provides a 
 buffer that will pass the produced event back to the Consumed Event Service.
 A consumed Own Event still only has one entry in the Event Table.  If the Event
-Variables are left as 0 or NULL, then the Consume Events Service will do nothing.
-If the Event Variables are populated as shown in the table in the Consumed Events
-Section above, the LEDs will behave accordingly to a Produced Event.
+Variables 2 onwards are left as 0 or NULL, then the Consume Events Service will
+do nothing.  Event Variable 1 will contain the value of the producer trigger 
+or switch. If the Event Variables are populated as shown in the table in the
+Consumed Events Section above, the LEDs will behave accordingly to a Produced Event.
 
 ### Node Variables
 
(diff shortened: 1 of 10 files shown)


dpharris:
How do we designate an event that is both Consumer AND Producer?

David

docs/VLCB4in4out_README.md
```diff
@@ -173,40 +191,7 @@ The FCU will show a new event produced by the Software Node in the VLCB4in4out e
...
-
-#### 'c'
-This character will return the CAN bus status.
-
-#### 'h'
-This character will return the event hash table.
-
-#### 'm'
-This character will return the amount of free memory. 
-
```
SvenRosvall, line 209: Oh, need to capture these in some other documentation file before they disappear from this file.

MartinDaCosta53:
@dpharris An event could reasonably be expected to be both a consumer and producer if it were a taught short event.  It could then be produced by more than one module and consumed by many.  EV1 is not looked at by a consumed event and so it does not know that it is a produced event as well.  If you populate EV2 onwards as if it were a consumed event, the event will be consumed.  If you have COE included then it that is how it will consume it's own event.  Hope this is clear.

docs/VLCB4in4out_README.md
MartinDaCosta53: Serial User Interface document now added with this information.

src/Controller.h
```diff
@@ -73,9 +73,7 @@ class Controller
   void indicateMode(byte mode);
   void indicateActivity();
   void setLearnMode(byte reqMode);
-  bool isSetProdEventTableFlag() { return setProdEventTable; }
-  void clearProdEventTableFlag();
```
SvenRosvall, line 77: Should you also get rid of the flag setProdEventTable ?

src/EventProducerService.cpp
```diff
@@ -23,173 +23,227 @@ void EventProducerService::setController(Controller *cntrl)
...
+      nn_en[0] = highByte(module_config->nodeNum);
+      nn_en[1] = lowByte(module_config->nodeNum); 
+    }
+    else
+    {
+      opCode = (state ? OPC_ACON1 : OPC_ACOF1);
+    }
+    
+    VlcbMessage msg;
+    msg.len = 5;
```
SvenRosvall, line 142: No, this should still be 6.

src/EventProducerService.cpp
```diff
@@ -23,173 +23,227 @@ void EventProducerService::setController(Controller *cntrl)
...
+  if ((uninit) && (module_config->currentMode == MODE_NORMAL))
   {
     setProducedEvents();
-    controller->clearProdEventTableFlag();
+    uninit = 0;
   }
 }
 
-void EventProducerService::sendEvent(bool state, byte index)
+void EventProducerService::createDefault(byte evValue)
```
SvenRosvall, line 65: Can we call this function createDefaultProducedEvent() ?
Just to be clear what it is about and what the single evValue is used for?

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -132,7 +132,7 @@ void setupVLCB()
   modconfig.EE_EVENTS_START = 50;
   modconfig.EE_MAX_EVENTS = 64;
   modconfig.EE_PRODUCED_EVENTS = NUM_SWITCHES;
-  modconfig.EE_NUM_EVS = NUM_LEDS;
+  modconfig.EE_NUM_EVS = 1 + NUM_LEDS;
```
SvenRosvall, line 135: Hmm, this is dangerous. Now the user sketch has to take the default produced event variable into account. What if the user forgets this?

src/Controller.h
MartinDaCosta53: Yes.  I'll update

src/EventProducerService.cpp
MartinDaCosta53: Your right.  My mistake.  Line 184 should be 7, as well.

src/EventProducerService.cpp
MartinDaCosta53: No problem, I'll update.

examples/VLCB_4in4out/VLCB_4in4out.ino
MartinDaCosta53: I also don't like having random numbers scattered about.  I was thinking of re-introducing the GLOBAL_EVS that we had at one stage in the CBUS version for handling SoD.  Alternatively, there could be a constant byte SYSTEM_EVS = 1.  Let's discuss later.

src/EventProducerService.cpp
```diff
@@ -23,173 +23,197 @@ void EventProducerService::setController(Controller *cntrl)
...
+      nn_en[0] = highByte(module_config->nodeNum);
+      nn_en[1] = lowByte(module_config->nodeNum); 
+    }
+    else
+    {
+      opCode = (state ? OPC_ACON3 : OPC_ACOF3);
+    }
+    
+    VlcbMessage msg;
+    msg.len = 5;
```
SvenRosvall, line 201: len should be 8 here.

dpharris:
What does  NUM_LEDS have to do with EE_NUM_EVS?
I would have thought maybe NVs?
A potential (ab)user wants to know!

David

examples/VLCB_4in4out/VLCB_4in4out.ino
SvenRosvall: Happy to leave this for now. We can improve this configuration later.

src/EventProducerService.cpp
MartinDaCosta53: Sorry, missed it! Now sorted

examples/VLCB_4in4out/VLCB_4in4out.ino
MartinDaCosta53: OK.  I'll sort it next time round and certainly before we go to a wider audience

dpharris:
Clear as mud ;-)

(1) I think we are allowing 'spoofing', so it could be short or long.
(2) I would have thought that the library would not deliver produced events
to the user, but only consumed events.  To do that it would check EV1, and
if zero supply, and if not not supply.  However, maybe it is ok to tell the
user, you are receiving all events in the event table. If they arrive, the
user can decide on what to do.  We are not deciding that EV1!=0 is only
useful for produced events, are we?  I would like to use it as an action
for consumed events, too.  Eg determine the channel for a LED, or set the
blink rate for my attached LED.

So the user API:
   void callback would be  userProcessEvent( state, index, msg)?
   void sendEvent( state, index )?
   void findEv( evn, OP, value, index)?
   optional: processMsg( msg )?

David

MartinDaCosta53:
@dpharris An EV is assigned to each LED to say what it does when an event is received to consume.  In the case of VLCB4in4out that will be nothing, or turn on/off or flash at one of two rates.
The NVs determine switch behaviour.  A toggle switch, on only, off only, push button.

dpharris:
@MartinDaCosta53 Is that a set function of your code?  Ie built into the LED class?  I would have thought leave that up the user.  They can decide what EV1 is used for, and if they want an LED, then:
   if( getEv(index, 1) == 5 ) {
    leds[5].setFlash(500);
  }
or sommat.  
David

dpharris:
@Martin Da Costa ***@***.***> I am showing my ignorance
:-).   So that is just a function of that sketch for VLCB4in4out, and not a
general rule?
David

MartinDaCosta53:
@dpharris It's an example.  Users are free to adapt it to do whatever they want.  By changing the argument in flash(), you can make the flash rate anything you like in mS.

dpharris:
Yes

PR #61, opened 2023-11-10
Introduce Minimum Node Service Documentation

MartinDaCosta53:


docs/MinimumNode.md
```diff
@@ -0,0 +1,151 @@
+# Minimum Node Service
```
SvenRosvall, line 1: Please name this file MinimumNodeService.md as this is the name of the class.

SvenRosvall:
Hmm. Sorry to be negative here. Most of this file repeats what is in the MNS specification. So I don't get what the purpose of this documentation is.

MartinDaCosta53:
Yes, I realised last evening that one bit that I'd missed out was probably the most important being how to configure it into a sketch!  Regarding the rest, the spec doesn't say how to implement the service.  This was written without reference to the spec but only the code as written.  The hope was that it would remove any ambiguity between the spec and actual implementation.  I do believe that we need something by way of a readme for each service.  If you think that this is too heavy or missing the mark, then I'm happy to simplify.

docs/MinimumNode.md
MartinDaCosta53: Done

docs/MinimumNodeService.md
```diff
@@ -0,0 +1,174 @@
...
+  virtual void setController(Controller *cntrl) override;
+  virtual void process(UserInterface::RequestedAction requestedAction) override; 
+  virtual Processed handleMessage(unsigned int opc, VlcbMessage *msg) override;
+
+  virtual byte getServiceID() override { return SERVICE_ID_MNS; }
+  virtual byte getServiceVersionID() override { return 1; }
+  
+  virtual void begin() override;
+  
+  // backdoors for testing
```
SvenRosvall, line 20: Please take out these lines that are used for testing. You don't describe them.

docs/MinimumNodeService.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Request Diagnostic Data - OPC_RDGN(0x87)
+
+If the Service Index is zero, then the module will send OPC_DGN(0xAC) with the number of 
+services supported.  It will then send an OPC_DGN for each of those services.
+
+If the Service Index is greater than zero, then a single OPC_DGN(0xAC) will be sent relating
+to that specific service.
+
+## User Sketch
```
SvenRosvall, line 142: This section should be moved to a separate page about writing sketches using the VLCB library. This section isn't specific to the MinimumNodeService.

We can move this section later.

SvenRosvall (approved):

docs/MinimumNodeService.md
MartinDaCosta53: Do you mean all of the lines or just the ones regarding backdoors?

docs/MinimumNodeService.md
MartinDaCosta53: OK

docs/MinimumNodeService.md
SvenRosvall: Just the backdoor ones. Exactly as you did it.

PR #64, opened 2023-11-16
Simplify teaching produced events

MartinDaCosta53:


examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -219,64 +223,89 @@ void loop()
...
+    for (byte i = 0; i < NUM_SWITCHES; i++) 
+    {
+      moduleSwitch[i].update();
+      if (moduleSwitch[i].changed())
+      {      
+        byte nv = i+1;
+        byte nvval = modconfig.readNV(nv);
+        bool state;
+
+        // DEBUG_PRINT(F("sk> Button ") << i << F(" state change detected. NV Value = ") << nvval);
```
SvenRosvall, line 265: What does "sk" mean? 
I guess it is short for "sketch". Please document this somewhere.

src/EventProducerService.cpp
```diff
@@ -62,16 +56,16 @@ void EventProducerService::process(UserInterface::RequestedAction requestedActio
...
   byte nn_en[4];
-  byte index;
   byte opCode;
-  index = module_config->findExistingEventByEv(1, evValue);
+
   if (index < module_config->EE_MAX_EVENTS)
   {
     module_config->readEvent(index, nn_en);
-    unsigned int nn = ((nn_en[0] << 8) && nn_en[1]);
+    unsigned int nn = ((nn_en[0] << 8) + nn_en[1]);
```
SvenRosvall, line 67: Please note that I changed this recently to use an if statement like "if ((nn_en[0] == 0) && (nn_en[1] == 0))"
This actually saves a few bytes.
nn is only used in the if-statement below (and the debug but that one can use the long expression).
You may get a merge conflict on this change.

examples/VLCB_4in4out/VLCB_4in4out.ino
MartinDaCosta53: Yes, it stands for sketch.  I found that when I was using debug it was difficult to track where they had come from.  Hence eps, ecs, sk, etc.  I'll add a note to the readme.

src/EventProducerService.cpp
MartinDaCosta53: That's simpler.  I'll change it.

src/EventTeachingService.cpp
```diff
@@ -371,70 +378,79 @@ Processed EventTeachingService::handleLearnEvent(VlcbMessage *msg, unsigned int
...
+            //DEBUG_SERIAL << F("ets> Update candidate") << endl;
+            unsigned long timeStart = millis();
+            bool terminate = false;
+            while ((timeOut > millis() - timeStart) && !terminate)
+            {
+              index = (byte)(*checkProduced)();
+              if (index < module_config->EE_PRODUCED_EVENTS)
+              {
+                // DEBUG_SERIAL << F("ets> Update index ") << index << endl;
+                terminate = true;
```
SvenRosvall, line 403: You can use "break;" here and get rid of the variable "terminate".

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -219,64 +223,89 @@ void loop()
...
-            state = (switchState[i]);
-            //DEBUG_PRINT(F("> Button ") << i << (moduleSwitch[i].fell() ? F(" pressed, send state: ") : F(" released, send state: ")) << state);
+byte checkProduced(void)
+{
+  check = true;
+  byte pIndex = processSwitches(check);
+  return pIndex;
+}
+
+byte processSwitches(bool check) {
```
SvenRosvall, line 239: This function does two things. 1) Check which switch has been changed and returns an index; 2) checks which swich has been changed (same so far) and sends an event. Note that for 2) it doesn't return anything eventhough it has a return type. 
Looks like these two things should be different functions. If you need to do one or another, please create a third function that calls either of the two functions. Keep the functions small.

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -219,64 +223,89 @@ void loop()
...
-        case 3:
-          // Toggle button
-          if (moduleSwitch[i].fell()) {
-            switchState[i] = !switchState[i];
-            state = (switchState[i]);
-            //DEBUG_PRINT(F("> Button ") << i << (moduleSwitch[i].fell() ? F(" pressed, send state: ") : F(" released, send state: ")) << state);
+byte checkProduced(void)
+{
+  check = true;
+  byte pIndex = processSwitches(check);
```
SvenRosvall, line 235: Call processSwitch(true) instead of using the global variable "check".
Get rid of the global variable "check".

src/EventTeachingService.cpp
MartinDaCosta53: Agreed.  I'll change it.

src/EventTeachingService.h
```diff
@@ -28,8 +28,10 @@ class EventTeachingService : public Service
 private:
   Controller *controller;
   Configuration *module_config;  // Shortcut to reduce indirection code.
+  byte (*checkProduced)();
```
SvenRosvall, line 31: You need to describe what checkProduced() is used for. 
If I read the code correct and based on our chat on Tuesday it looks like this call-back function is used when FCU is teaching an event and it needs to know which input this event is for.
Maybe call it checkInputProduced() ?

src/EventTeachingService.h
```diff
@@ -28,8 +28,10 @@ class EventTeachingService : public Service
...
 
   bool bLearn = false;
+  unsigned long timeOut = 5000;
```
SvenRosvall, line 34: Need a comment to clarify what this is a timeout for.
Also, as it is never changed I would make it "const" and give it a name with all upper case letters.

src/EventTeachingService.cpp
```diff
@@ -371,70 +378,79 @@ Processed EventTeachingService::handleLearnEvent(VlcbMessage *msg, unsigned int
...
-          module_config->updateEvHashEntry(index);
-          return PROCESSED;
+          // Are we a producer and is this a produced event to update?
+          // Wait timeOut seconds to see if producer button pressed.
+          if ((controller->getParam(PAR_FLAGS) & PF_PRODUCER) && (evindex ==1))       
+          {
+            //DEBUG_SERIAL << F("ets> Update candidate") << endl;
+            unsigned long timeStart = millis();
+            bool terminate = false;
+            while ((timeOut > millis() - timeStart) && !terminate)
```
SvenRosvall, line 397: This loop will block processing for 5 seconds. No incoming events will be received. No heartbeats or any other processing.
Might be tricky to refactor out this. We can look at it later to get this change through.

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -219,64 +223,89 @@ void loop()
...
+    
+  if (check)
+  {
+    for (byte i = 0; i < NUM_SWITCHES; i++) 
+    {
+      moduleSwitch[i].update();      
+      if (moduleSwitch[i].changed())
+      {
+        //DEBUG_PRINT(F("sk> Check index is ") << i);
+        return i;
```
SvenRosvall, line 249: You don't check the NVs here for what is happening on each input happening. Say that you only have an event on closing a switch. This code will also trigger on breaking the switch.

examples/VLCB_4in4out/VLCB_4in4out.ino
MartinDaCosta53: Oh so simple!  You wouldn't believe the trouble I had thinking this one through!

docs/VLCB4in4out_README.md
```diff
@@ -80,18 +79,21 @@ OP_CODE | HEX | Function
...
-### Event Variables
+## Set Up
+
+The module is initialised by pressing the VLCB button (formerly CBUS button) for
+6 seconds when the green LED goes off and the Yellow LED flashes. A module name and
+Node Number can then be set via the FCU in the normal manner.
 
-Event Variable 1 (EV1) is reserved for system use to identify the source of
-produced event.
+## Event Variables
```
SvenRosvall, line 92: I don't think you need a heading for "Event Variables". The sub-section on produced events don't mention event variables. Maybe call this heading "Events" only?

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -219,64 +223,89 @@ void loop()
   }
 
   // test for switch input
-  processSwitches();
+  check = false;
+  processSwitches(check);
```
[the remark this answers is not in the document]
MartinDaCosta53: Yes, I've reverted to the way it was.

examples/VLCB_4in4out/VLCB_4in4out.ino
MartinDaCosta53: Yes, done that

src/EventProducerService.cpp
```diff
@@ -62,17 +56,16 @@ void EventProducerService::process(UserInterface::RequestedAction requestedActio
   }
 }
 
-void EventProducerService::sendEvent(bool state, byte evValue)
+void EventProducerService::sendEvent(bool state, byte index)
```
SvenRosvall, line 59: You changed the parameter name "evValue" to "index" here. I guess you should do the same for the other sendEvent functions.

src/EventTeachingService.h
MartinDaCosta53: OK.

src/EventTeachingService.cpp
MartinDaCosta53: Yes, blocking gave me a headache.  However, It is 5 seconds and we are in LEARN_MODE, so I figured it would be OK as little else can happen.

docs/VLCB4in4out_README.md
MartinDaCosta53: Good point.

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -224,31 +228,47 @@ void loop()
...
+      if (moduleSwitch[i].changed())
+      {
+        //DEBUG_PRINT(F("sk> Check index is ") << i);
+        return i;
+      } 
+    }      
+      return 0xFF;   
+}
+
+byte processSwitches(void) {
```
SvenRosvall, line 245: Please revert the return type "byte" back to "void".

examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -115,11 +115,12 @@ Bounce moduleSwitch[NUM_SWITCHES];  //  switch as input
 LEDControl moduleLED[NUM_LEDS];     //  LED as output
 byte switchState[NUM_SWITCHES];
 
+bool check;
```
SvenRosvall, line 118: You don't need this anymore.

PR #65, opened 2023-11-18
Update tests and code cleanups after updates to teaching service.

SvenRosvall:


MartinDaCosta53 (approved):
This all seems very clear and makes it much easier to read.  Sorry about the { at line 383.  I left these in while I was developing and forgot to take them out at the end.

PR #66, opened 2023-11-21
Add CMake configuration to build without the Arduino IDE

chrisa:
This allows VLCB sketches to be built entirely with cmake, without the Arduino IDE but still using the Arduino SDK.

- tests may be built with Clang's sanitizers, and pass cleanly.
- testAll exits non-zero on failure, for use with ctest.
- all example sketches are buildable with CMake.

See `build-all.sh` for cmake command lines.


dpharris:
Nice. I am just watching the CLion intro video... it looks sweet.

test/testAll.cpp
```diff
@@ -67,5 +69,5 @@ int main(int argc, const char * const * argv)
       }
     }
   }
-  return 0;
+  return failures;
```
SvenRosvall, line 72: Would be nice to print out a message with the failure count if there are any failures.

PR #67, opened 2023-11-24
Fixes for event teaching where application doesn't provide a checkInputProduced callback.

SvenRosvall:


src/EventTeachingService.cpp
MartinDaCosta53, line 1: I assume that != takes precedence over &&, but would it aid readability if 
checkInputProduced != nullptr
where in parentheses?

SvenRosvall: Indeed!

PR #74, opened 2023-12-08
Remove CMDERR from checks for short messages.

SvenRosvall:


MartinDaCosta53:
This change aligns with the specification requirement.

PR #75, opened 2023-12-08
Refactor Configuration::findExistingEvent

SvenRosvall:
Simplify lookup for existing events.


MartinDaCosta53:
Does the hash table help us?  Why not just compare nn with the events table and if there is a match compare the en?

PR #76, opened 2023-12-17
Revert to using EV1 for producer ident

MartinDaCosta53:
Removes need for delay in teaching whilst avoiding the need to use the FCU Debug facility.  All teaching can be done using FCU default methods.


fletchjp:
That looks good. It seems a good thing that I have waited until now to tackle event learning with my VLCB examples.

src/EventProducerService.cpp
```diff
@@ -90,22 +118,29 @@ void EventProducerService::sendEvent(bool state, byte index)
...
   if (index < module_config->EE_MAX_EVENTS)
   {
     byte opCode;
     byte nn_en[4];
     module_config->readEvent(index, nn_en);
+    //DEBUG_SERIAL << F("eps>index = ") << index << F(" , Node Number = 0x") << _HEX(nn_en[0]) << _HEX(nn_en[1]) << endl;
     if ((nn_en[0] == 0) && (nn_en[1] == 0))
     {
-      opCode = (state ? OPC_ASON1 : OPC_ASOF1);
+      opCode = (state ? OPC_ASON : OPC_ASOF);
```
SvenRosvall, line 132: Shouldn't these opcodes be OPC_ASON1 and OPC_ASOF1 here? We are sending one byte of data.
I guess this is a copy paste error.
See also similar lines below.

MartinDaCosta53: Well caught!  Now corrected through 1, 2 & 3.

docs/VLCB4in4out_README.md
```diff
@@ -83,21 +84,21 @@ OP_CODE | HEX | Function
...
-## Set Up
+## Events
 
-The module is initialised by pressing the VLCB button (formerly CBUS button) for
-6 seconds when the green LED goes off and the Yellow LED flashes. A module name and
-Node Number can then be set via the FCU in the normal manner.
+### Event Variables
 
-## Events
+Event Variable 1 (EV1) identifies a consumed only event if its value is zero.
```
SvenRosvall, line 91: Is this sentence correct. Isn't consumed events controlled by EV2 and upwards?

docs/VLCB4in4out_README.md
```diff
@@ -114,27 +115,36 @@ The following EV values are defined to control the LEDs in this example:
...
 
 If the Produced Events Service and the Consume Events Service are both applied,
 the Consume Own Events Service can also be enabled.  This service provides a 
 buffer that will pass the produced event back to the Consumed Event Service.
-A consumed Own Event still only has one entry in the Event Table.  If the Event
-Variables are left as 0 or NULL, then the Consume Events Service will do nothing.
-If the Event Variables are populated as shown in the table in the Consumed Events
-Section above, the LEDs will behave accordingly to a Produced Event.
+A Consume Own Event still only has one entry in the Event Table.  If the Event
+Variables 2 onwards are left as 0 or NULL or any undefined value, then the
```
SvenRosvall, line 142: Can an event variable be set to "NULL" ?

SvenRosvall:
Please have a look at VLCB_1in1out.ino and the other examples. They may need some adjustments too.

docs/VLCB4in4out_README.md
```diff
@@ -148,12 +158,27 @@ The following NV values define input switch function:
 
 NV Value | Function
 --------|--------
- 0 | On/Off switch
- 1 | On only push button
- 2 | Off only push button
- 3 | On/Off toggle push button
+ 0 | Do nothing
```
SvenRosvall, line 161: Yes, this is a good change. How did this work before?

src/EventProducerService.cpp
```diff
@@ -38,37 +38,65 @@ void EventProducerService::setProducedEvents()
...
+    //DEBUG_SERIAL << F("eps>index = ") << index << F(" , Node Number = 0x") << _HEX(nn_en[0]) << _HEX(nn_en[1]) << endl;
     if ((nn_en[0] == 0) && (nn_en[1] == 0))
     {
       opCode = (state ? OPC_ASON : OPC_ASOF);
       nn_en[0] = highByte(module_config->nodeNum);
       nn_en[1] = lowByte(module_config->nodeNum); 
     }
+    else if ((nn_en[0] == 0xff) && (nn_en[1] == 0xff))
+    {
+      recreateDefaultEvent(evValue);
```
SvenRosvall, line 97: This looks strange to me. I find it strange to recreate devault events while sending events. 
As I read the code, if the event we are about to send is NN/EN == FFFF (Event is found but not set up(?)) then we recreate the default events. Is this some way to catch a bad configuration? 
Under what conditions whould you enter into this code here? How can we have an event in the table but with unset NN/EN?

src/EventTeachingService.cpp
```diff
@@ -399,37 +393,31 @@ Processed EventTeachingService::handleLearnEvent(VlcbMessage *msg, unsigned int
...
+        {
+          module_config->writeEvent(index, &msg->data[1]);
+          // no need to write eventEV as, by definition, it hasn't changed
+          // recreate event hash table entry
+          module_config->updateEvHashEntry(index);
+          return PROCESSED;
+        }
+      }     
+      
+      // search for this NN, EN as we may just be adding an EV to an existing learned event
```
SvenRosvall, line 411: Please fix indentation here.

src/EventTeachingService.cpp
```diff
@@ -399,37 +393,31 @@ Processed EventTeachingService::handleLearnEvent(VlcbMessage *msg, unsigned int
...
+  if ((evindex == 1) && (evval > 0))
+      {
+        byte index = module_config->findExistingEventByEv(evindex, evval);
+        if (index < module_config->EE_MAX_EVENTS)
+        {
+          module_config->writeEvent(index, &msg->data[1]);
+          // no need to write eventEV as, by definition, it hasn't changed
+          // recreate event hash table entry
+          module_config->updateEvHashEntry(index);
+          return PROCESSED;
```
SvenRosvall, line 407: Shouldn't there be response codes here? Old code had both WRACK and GRSP(OK).

docs/VLCB4in4out_README.md
MartinDaCosta53: That's true but the sentence is correct in that if EV1 is zero then the event is consumed only.  If it is non-zero, then it is a produced event and possibly consumed as well if EV2 onwards are non-zero.

docs/VLCB4in4out_README.md
MartinDaCosta53: No, I don't think they can.  I'll update.

docs/VLCB4in4out_README.md
MartinDaCosta53: If it was undefined then it would do nothing.  However, that wasn't documented and so I made the change.  This is important because if you now set the NV to 0 or an unspecified number, a send request is not generated and so a deleted default produced event isn't re-instated.  You need to set the NV to a defined value to allow a deleted default produced event to be re-instated.

src/EventProducerService.cpp
MartinDaCosta53: If you delete a default event, the nn and en values in the events table get set to 0xFF.  However, EVs are not deleted and so if that index position hasn't been overwritten by another teach event the pressing of a switch to re-instate will find this index location with a nn_en of 0xFFFF.  Thinking about it, if that index position has now been re-used, the findExistingEventbyEv() will return an index > EE_MAX_EVENTS and so I need an else statement to call recreateDefaultEvent at the end of the first if statement as well.

src/EventTeachingService.cpp
MartinDaCosta53: OK

src/EventTeachingService.cpp
MartinDaCosta53: Yes, you're probably right.  I'll add it in.

src/EventProducerService.cpp
SvenRosvall: Yes, I was thinking that the entry in evhashtbl would be cleared which means it wouldn't be found.
Anyway, I think sendEvent() is the wrong place to recreate default events. If you have deleted such an event then you don't want it recreated. I agree that in that case you shouldn't call sendEvent with that EV1 value. Doing so should be considered an error, not recreating the events.
We need another mechanism for recreating the default events. Keep in mind that Ian wants the NNCLR as a way to get rid of all default events. ;) For now NNRSM seems to be the only valid way to get default events back.
Happy to discuss this further.

src/EventTeachingService.cpp
SvenRosvall: Indentation is still out for lines 396-427

src/EventProducerService.cpp
MartinDaCosta53: If the NV for a switch (in the case of 4in4out) has a valid setting i.e. 1-4, then pressing the switch will re-instate the default.  This is done using sendEvent as the simplest way.  If the NV is set to 0 or an undefined number (5 upwards in the case of 4in4out) then sendEvent is not sent.  The problem with using the NNRSM route is that you delete everything else that may have been taught!  NNCLR gets rid of all events, not just defaults.  If the NVs are set to 0, then they will not be re-instated.

src/EventTeachingService.cpp
MartinDaCosta53: Sorted

src/EventProducerService.cpp
SvenRosvall: Reading code better now and see that you are only reinstating _one_ event. 
And your comment on the NV controlling the existence of the produced event makes sense.

MartinDaCosta53: It's good to be challenged on these things as it makes me think about what I've done.  It also helped me pick up an omission that would have become a head-scratching issue in the future.

src/EventProducerService.cpp
```diff
@@ -38,37 +38,65 @@ void EventProducerService::setProducedEvents()
...
+      index = module_config->findEventSpace();
+    }
+        
     module_config->writeEvent(index, data);
+    module_config->writeEventEV(index, 1, i);
     module_config->updateEvHashEntry(index);    
   }    
 }
 
+void EventProducerService::recreateDefaultEvent(byte evValue)
```
SvenRosvall, line 54: I see code duplication here. Can you let setProducedEvents() call recreateDefaultEvent() in the loop. 
Does the condition there matter? For the conditions where is called (changing for Uninitialized to Normal), I expect the event table to be empty. I.e. findExistingEventByEv() won't find the events which means that findEventSpace() will be called, which is what recreateDefaultEvent() uses.
We can leave this for a later change if you prefer to isolate that refactoring and do more testing for just this change.

examples/VLCB_1in1out/VLCB_1in1out.ino
SvenRosvall, line 1: This sketch still has checkInputProduced().

MartinDaCosta53: Sorry.  Now fixed

PR #79, opened 2023-12-20
Send message when default event created.

MartinDaCosta53:
Sending an event tells a configuration tool of the events existence.

### DIFF
--- a/src/EventProducerService.cpp
+++ b/src/EventProducerService.cpp
@@ -33,21 +33,22 @@ void EventProducerService::setProducedEvents()
 { 
   for (byte i = 1; i <= module_config->EE_PRODUCED_EVENTS; i++)
   {
-    createDefaultEvent(i);    
+    createDefaultEvent(i);
+    opCode = OPC_ACOF;
+    generateMessage();
   }    
 }
 
 void EventProducerService::createDefaultEvent(byte evValue)
 {
-  byte data[4];
-  data[0] = highByte(module_config->nodeNum);
-  data[1] = lowByte(module_config->nodeNum);
-  data[2] = 0;
-  data[3] = evValue;
+  nn_en[0] = highByte(module_config->nodeNum);
+  nn_en[1] = lowByte(module_config->nodeNum);
+  nn_en[2] = 0;
+  nn_en[3] = evValue;
   
   byte index = module_config->findEventSpace();
   
-  module_config->writeEvent(index, data);
+  module_config->writeEvent(index, nn_en);
   module_config->writeEventEV(index, 1, evValue);
   module_config->updateEvHashEntry(index);
 }
@@ -66,10 +67,13 @@ void EventProducerService::process(UserInterface::RequestedAction requestedActio
 void EventProducerService::sendEvent(bool state, byte evValue)
 {
   byte index = module_config->findExistingEventByEv(1, evValue);
-  if (index < module_config->EE_MAX_EVENTS)
+  
+  if (index >= module_config->EE_MAX_EVENTS)
   {
-    byte opCode;
-    byte nn_en[4];
+    createDefaultEvent(evValue);
+  }
+  else
+  {    
     module_config->readEvent(index, nn_en);
     //DEBUG_SERIAL << F("eps>index = ") << index << F(" , Node Number = 0x") << _HEX(nn_en[0]) << _HEX(nn_en[1]) << endl;
     if ((nn_en[0] == 0) && (nn_en[1] == 0))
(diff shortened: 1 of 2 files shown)


src/EventProducerService.cpp
```diff
@@ -78,43 +82,28 @@ void EventProducerService::sendEvent(bool state, byte evValue)
...
-    if (coeService)
-    {
-      coeService->put(&msg);
-    }
-  }
-  else
-  {
-    createDefaultEvent(evValue);
   }
+  generateMessage();
```
SvenRosvall, line 94: Hmm, this doesn't look right. What if the if-statement on line 71 is true? Then this line will generate a message where opcode and nn_en are not set.
If you used function parameters instead of class variables then this would be easier to spot.

src/EventProducerService.h
```diff
@@ -33,13 +33,19 @@ class EventProducerService : public Service {
...
+  void generateMessage(byte data1, byte data2, byte data3);
 
 private:
   Controller *controller;
   Configuration *module_config;  // Shortcut to reduce indirection code.
   ConsumeOwnEventsService *coeService;
   void (*eventhandler)(byte index, VlcbMessage *msg);
 
+  byte opCode;
+  byte nn_en[4];
```
SvenRosvall, line 48: Hmm, this doesn't look right. This data belongs to method calls and do not reflect the state of the EPService.
Yes, it is a technique used sometimes to save a few bytes, but I find it confusing.

MartinDaCosta53: I did this to keep them in scope across the various methods that use them.  I can look again at using a lot of arguments in the various calls instead.

src/EventProducerService.cpp
MartinDaCosta53: nn_en will have been set by the createDefaultEvent(), but you're right about the op-code being undefined.

SvenRosvall: Oh, relying on such side effects is scary.

src/EventProducerService.cpp
```diff
@@ -33,21 +33,22 @@ void EventProducerService::setProducedEvents()
 { 
   for (byte i = 1; i <= module_config->EE_PRODUCED_EVENTS; i++)
   {
-    createDefaultEvent(i);    
+    createDefaultEvent(i);
+    opCode = OPC_ACOF;
+    generateMessage();
```
SvenRosvall, line 38: I don't think we should generate messages here as no input has changed yet. No existing CBUS modules do this. 
And are you sure the input state is OFF? :)

src/EventProducerService.cpp
MartinDaCosta53: It will be set at line 91

MartinDaCosta53: This actually turned out to require more thinking than I'd expected.  Moving the declaration of nn_en and opCode to the .h file so that they stayed in scope across a number of methods made it easier.  I can think further about it and use arguments in the calls instead if that gives better code.

src/EventProducerService.cpp
MartinDaCosta53: The message is arbitrary.  I used off so as not to leave anything on!  If you don't do this, then you won't see the defaults in the FCU until you read events in the nodes menu.  It was an after-thought in doing this but as valid as sending a message when restoring a default.  I don't consider what existing CBUS modules may or may not do as being a good guide.  It can come out easily enough if you feel strongly about it.

src/EventProducerService.cpp
SvenRosvall: Line 91 is in the false branch of the if-statement on line 71, thus it will not be executed if that condition is true.

src/EventProducerService.cpp
SvenRosvall: I still don't like these generated events.
1) As you say the off state is arbitrary, and may indicate the wrong state. Shouldn't matter as these events won't be taught to any other nodes at this stage. But it just doesn't feel right.
2) Say the user doesn't want a few of these events and changes their NVs. If these events are created before the user gets a chance to change these NVs, then those extra events will show up in FCU anyway. And we know how messy it is to delete events.
3) As you have seen before, I do use existing CBUS modules as inspiration. What works well for them should be kept. If we want to do something different we have to consider any other side effects.

MartinDaCosta53: 1. It seems logical that a state should be initialised off.
2. They're produced upon initialisation and so are there anyway even if they can't yet be seen in the FCU.  However, as soon a you read the events, it will appear in the FCU even if the NV is 0.  Now, we could not create any defaults until the event tries to be produced, a bit like re-instating a default previously deleted.  However, this could get tedious if there are a lot.  On the other hand, maybe if there are a lot as in Ian's Universal, it should have a dedicated service.
3. I also take inspiration from existing CBUS modules but that should not constrain us.

src/EventProducerService.cpp
MartinDaCosta53: No it isn't.  It is always executed if the nn != 0xFFFF

src/EventProducerService.h
```diff
@@ -33,13 +33,19 @@ class EventProducerService : public Service {
   void sendEvent(bool state, byte evValue, byte data1);
   void sendEvent(bool state, byte evValue, byte data1, byte data2);
   void sendEvent(bool state, byte evValue, byte data1, byte data2, byte data3);
+  void generateMessage();
+  void generateMessage(byte data1);
+  void generateMessage(byte data1, byte data2);
+  void generateMessage(byte data1, byte data2, byte data3);
```
SvenRosvall, line 39: These generateMessage() functions should be private. You don't want anyone call these when they might not understand that they need to set opcode and nn_en first.

MartinDaCosta53: OK

src/EventProducerService.cpp
SvenRosvall: The else branch for the if statement on line 71 spans lines 76-92. It is tricky with all the nested if statements but as I read the code line 91 will not be executed if the condition on line 71 is false.

MartinDaCosta53: Ah, yes, I see what you mean.  I've duplicated line 91 after line 73 to resolve.  I can't move line 91 or I'll upset the short event op-codes.

SvenRosvall:
Implemented another way.

dpharris:
Can you expand on 'another way'?

SvenRosvall:
See actual implementation in https://github.com/SvenRosvall/VLCB-Arduino/pull/80

PR #81, opened 2023-12-29
Refactor factory reset.

SvenRosvall:
Change how a push-button on module start causes a factory reset. Move the code for checking the pushbutton from Configuration to LEDUserInterface. Clean-up UserInterface to remove virtual functions that are only used for handling factory reset.

### DIFF
--- a/examples/VLCB_1in1out/VLCB_1in1out.ino
+++ b/examples/VLCB_1in1out/VLCB_1in1out.ino
@@ -110,7 +110,7 @@ void setupVLCB()
   controller.setName(mname);
 
   // module reset - if switch is depressed at startup and module is in Uninitialised mode
-  if (ledUserInterface.isButtonPressedForReset(modconfig.currentMode))
+  if (ledUserInterface.isButtonPressed())
   {
     Serial << F("> switch was pressed at startup in Uninitialised mode") << endl;
     modconfig.resetModule();
--- a/examples/VLCB_1in1out/VLCB_1in1out.ino
+++ b/examples/VLCB_1in1out/VLCB_1in1out.ino
@@ -110,11 +110,11 @@ void setupVLCB()
   controller.setName(mname);
 
   // module reset - if switch is depressed at startup and module is in Uninitialised mode
-//  if (userInterface.isButtonPressed() && modconfig.currentMode == MODE_UNINITIALISED)
-//  {
-//    Serial << F("> switch was pressed at startup in Uninitialised mode") << endl;
-//    modconfig.resetModule(&userInterface);
-//  }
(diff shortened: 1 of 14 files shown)


MartinDaCosta53 (commented):
Holding the button pressed at start-up will only reset the unit if it is unconfigured.  Do we want this restriction?  Is there a need to reset an unconfigured unit?  Surely this should work whatever mode the module is in.

MartinDaCosta53 (commented):
The existing double press of the button to reset is Duncan's original method.  I would suggest that it is not now needed as reset can be instigated from button hold at start-up or by using OPC_NNRSM.  A crashed unit will need to be start-up cycled by power off / on or pressing the reset button.  In either case, the VLCB button can be held down if we want to wipe the module.

MartinDaCosta53:
> > The existing double press of the button to reset is Duncan's original method.
> > Do you mean pressing the 'z' key twice?
> 
> The double 'z' key press for reset was in the SerialUserInterface. But it is gone now. I have a memory that you were the one to remove that. I can't find it in the Git history.

I added the 'z' key twice to Duncan's original SerialUserInterface.  I left it out when we migrated to VLCB, which is why you can't find it in GitHub history.
What I was referring to is the pressing of the VLCB button twice for 5 secs within 30 secs. (Configuration.cpp lines 503 - 541.

SvenRosvall:
> What I was referring to is the pressing of the VLCB button twice for 5 secs within 30 secs. (Configuration.cpp lines 503 - 541.

Hmm, are we reading this code differently? Lines 503-541 refer to Configuration::resetModule(UI) which I have replaced with LEDUserInterface::isButtonPressedForReset(). Don't confuse this with Configuration::resetModule() without the UI parameter.

Configuration::resetModule(UI) was only called only from the example sketches during startup. (in the setup()) 
I read this code carefully again and see what you mean. It does a reset if the first press is at least 5 seconds. I did not realise that if you release the button within the 5 seconds you can press it again for 5 seconds. This looks unnescessarilly complicated. (And I misread the code.)

Now that you highlighted the original behaviour I realise I can clean up LEDUserInterface::isButtonPressedForReset() a bit more. And the comments are misleading.

So are you saying that you are happy with just the 5 second button press at start-up of the module?

MartinDaCosta53:
Once again, this was inherited from Duncan's CBUS code.  The way I read it is that if you press the button for 5 secs and then press it again for 5 secs within 30 secs, you perform a resetModule.  I think that just holding the button down at startup or module reboot is adequate.  The module reboot would come from pressing the Arduino reset button.

dpharris:
I do not think we can assume a rest button. I do not really have a
preference other than the instructions on use are clear.  I would prefer a
long press (>10 sec) to set a rest-to-manu-settings, which sets indicators,
with perhaps a confirmatory press.  But will adapt to whatever you decide.

David

PR #82, opened 2024-01-17
Remove bModeSetup from MinimumNodeService.

SvenRosvall:
Replace with checks for instantMode.


MartinDaCosta53:
That is much tidier.

PR #85, opened 2024-01-23
Complete update for Action (Command) bus

MartinDaCosta53:


examples/VLCB_4in4out/VLCB_4in4out.ino
```diff
@@ -58,6 +58,8 @@
 // VLCB library header files
 #include <Controller.h>                   // Controller class
 #include <CAN2515.h>               // CAN controller
+#include <Switch.h>
+#include <LED.h>
```
SvenRosvall, line 62: Are these actually needed? You don't use the types Switch and LED here.

SvenRosvall: I commented out these two #includes from 1in1out and it compiles just fine.
I think these #includes are remains from Duncan's examples where we did use the types Switch and LED directly. But now we have the LEDUserInterface.
So I guess we should remove these #includes from all example sketches.

MartinDaCosta53:
You're quite right.  I've taken them out and it compiles and loads and runs perfectly well without these includes.  I'll close the PR.

PR #86, opened 2024-01-25
Fix potential problem with overwriting current command element on buffer overflow.

SvenRosvall:
Fix potential problem with overwriting current command element on buffer overflow.
Make CircularBuffer API work with references instead of pointers.


MartinDaCosta53:
This all looks good to me.  I like the simplification of using references instead of pointers.

PR #87, opened 2024-01-26
Simplify mode setting

SvenRosvall:
Improve test for default produced events.
Simplify mode setting so that node number is also set at the same time.


MartinDaCosta53 (approved):
No problems with this.

PR #88, opened 2024-01-26
Minor detail changes.

MartinDaCosta53:


docs/Design.md
```diff
@@ -5,19 +5,19 @@ This document describes the main components within the VLCB library and how they
...
-The user sketch can select the set of services needed to provide the functionality
-that is necessary for the VLCB module that is created.
+Services implement various groups of functionality within VLCB such as events or DCC control.
+The user sketch can select the set of services needed to provide the necessary functionality
+for the VLCB module that is created.
 There are also services for user interfaces and communication over different transports such as CAN, Wifi and BLE.
 Currently only a CAN transport using the CAN2515 chip is included in the library.
 
-The library supports a set of storage for node variables and event variables in EEPROM or Flash memory.
+The library supports storage for node variables and event variables in EEPROM or Flash memory.
```
SvenRosvall, line 20: Good, but I miss the bit about that different types of EEPROM storages are supported by the library..

docs/Design.md
```diff
@@ -44,29 +44,31 @@ from the controller is an outgoing message it sends it to the CAN bus.
...
 ### Dataflow
-Most of VLCB functionality use a message object ```VlcbMessage``` that is used to pass incoming and
+Most of the VLCB functionality uses a message object ```VlcbMessage``` that passes incoming and
 outgoing messages around via the Command bus. 
 The ```VlcbMessage``` object contains 8 bytes where the first is the op-code and the remaining 7 bytes
 are any optional data bytes for that op-code.
 
-The ```CanTransport``` object translates the VlcbMessage object to from an object that represents
-a CAN frame which contains an id, 8 bytes of data (same as the VlcbMessage) and the flags
+The ```CanTransport``` object translates the VlcbMessage object to form an object that represents
```
SvenRosvall, line 56: I actually meant "to and from".

MartinDaCosta53: OK.  We can do it like that and then remove the new line 66.  Do you want me to amend?

docs/Design.md
MartinDaCosta53: I think that this gets covered lower down in "Configuration".

docs/Design.md
MartinDaCosta53: I've changed to reflect your original intention.

SvenRosvall (approved):

PR #90, opened 2024-01-29
Example sketches with Manufacturer MANU_DEV.

SvenRosvall:
After recent discussions I think the examples here should have their manufacturer set to MANU_DEV as they are not released kits.


MartinDaCosta53:
Good move!

PR #92, opened 2024-01-31
Avoid copying parameter to CAN2515::begin().

SvenRosvall:
Pass SPI parameter by reference to avoid copying it on stack. 
Saves 14 bytes of program memory.


MartinDaCosta53:
Every little bit helps

PR #94, opened 2024-01-31
Functions for converting an int to/from two bytes.

SvenRosvall:
Simplifies code that is getting or setting two bytes from/to a given value.
I did this change to see if I could reduce program memory because of the code duplication. But unfortunately it doesn't save anything. But it is still useful as I think the code is easier to read.


MartinDaCosta53 (approved):
Jury is out on whether it's easier to read.  For me, if I didn't know what it was doing I might have been scratching my head for a few minutes.  It's certainly more compact. Pity about not saving program size.

PR #96, opened 2024-02-11
Omit UI in rqsd response

SvenRosvall:
As discussed last VLCB meeting


MartinDaCosta53:
Looks fine and is what we agreed last evening.

PR #98, opened 2024-02-24
Introduce Op-Code Consumer Service and add to vlcbdefs

MartinDaCosta53:


SvenRosvall:
I am not fully convinced about this service. I had a look at the sketch you sent me earlier that shows how you use this service. I would have created a separate class that implements the service interface. OK, there may be a complexity in writing such a service if the sketch writer is not aware of what an Action object is. 
A compromise might be to create the OpCodeConsumerService as an abstract base class where the user creates a child class that implements the virtual function opCodeHandler(). This avoids those pesky function pointers. And it also allows users to collect statistics and error counters within the service object.

PR #99, opened 2024-02-26
REVAL for EV0 shall return number of EVs for this event and values for all of those EVs.

SvenRosvall:


MartinDaCosta53 (commented):
EventTeaching.cpp line 318 //TODO....  The op-code spec says yes.  Do you want to change it now or later?

PR #100, opened 2024-02-28
Avoid duplication of NN-EN when creating default produced events

MartinDaCosta53:


src/EventProducerService.cpp
```diff
@@ -39,15 +39,32 @@ void EventProducerService::setProducedEvents()
...
+  unsigned int eventNum = 0;
+  unsigned int nodeNum = module_config->nodeNum;
   byte data[4];
-  Configuration::setTwoBytes(&data[0], module_config->nodeNum);
-  data[2] = 0;
-  data[3] = evValue;
   
   byte index = module_config->findEventSpace();
+  //eventNum = (int)index + 1;
+  for(eventNum = 1; eventNum <= module_config->EE_MAX_EVENTS; eventNum++)
```
SvenRosvall, line 48: Please put a space between "for" and "(".
Same on line 50.

src/EventProducerService.cpp
```diff
@@ -39,15 +39,32 @@ void EventProducerService::setProducedEvents()
...
   
+  Configuration::setTwoBytes(&data[0], nodeNum);
+  Configuration::setTwoBytes(&data[2], eventNum);
+   
   module_config->writeEvent(index, data);
   module_config->writeEventEV(index, 1, evValue);
+  
+  for (byte i = 2; i <= module_config->EE_NUM_EVS; i++)
+  {
+    module_config->writeEventEV(index, i, 0);
```
SvenRosvall, line 66: Good catch. Should have been done before!

src/EventProducerService.cpp
```diff
@@ -39,15 +39,32 @@ void EventProducerService::setProducedEvents()
 
 byte EventProducerService::createDefaultEvent(byte evValue)
 {
+  unsigned int eventNum = 0;
+  unsigned int nodeNum = module_config->nodeNum;
   byte data[4];
```
SvenRosvall, line 44: Please move this declaration of data down to before line 58 where it is first used.
Maybe also rename this to "nn_en" to make it clearer what it is for.

src/EventProducerService.cpp
MartinDaCosta53: I agree.  It solves a lot of aggravation with ev variables coming out as 0xFF

src/EventProducerService.cpp
```diff
@@ -39,15 +39,32 @@ void EventProducerService::setProducedEvents()
...
-  Configuration::setTwoBytes(&data[0], module_config->nodeNum);
-  data[2] = 0;
-  data[3] = evValue;
   
   byte index = module_config->findEventSpace();
```
SvenRosvall, line 46: What if the event table is full?
No we didn't check this before either so you don't have to fix that now. Just want to raise this issue.
This will be tricky to fix as the error condition needs to be worked back to sendEvent(). Should sendEvent() ignore this?

src/EventProducerService.cpp
MartinDaCosta53: I assume you only mean the declaration at line 43 to be moved as eventNum is used at line 48.
These are two separate variables.  Not sure what you mean by renaminh to "nn_en"

src/EventProducerService.cpp
MartinDaCosta53: Good point.  I need to think about this.  It clearly needs to flag an error.

src/EventProducerService.cpp
SvenRosvall: No, i meant "data" on line 44. "data" is a generic name so want to call it "nn_en".

PR #101, opened 2024-02-28
Add error codes for event table full

MartinDaCosta53:


src/EventProducerService.cpp
```diff
@@ -43,11 +43,12 @@ byte EventProducerService::createDefaultEvent(byte evValue)
   unsigned int nodeNum = module_config->nodeNum;
   
   byte index = module_config->findEventSpace();
-  //eventNum = (int)index + 1;
   for (eventNum = 1; eventNum <= module_config->EE_MAX_EVENTS; eventNum++)
   {
     if (module_config->findExistingEvent(nodeNum, eventNum) == module_config->EE_MAX_EVENTS)
     {
+      controller->sendCMDERR(CMDERR_TOO_MANY_EVENTS);
+      controller->sendGRSP(OPC_RQNPN, getServiceID(), CMDERR_INV_PARAM_IDX);
```
SvenRosvall, line 51: Shouldn't the GRSP have the same error code? I.e. CMDERR_TOO_MANY_EVENTS

SvenRosvall, line 51: What happens next? eventNum will be set to EE_MAX_EVENTS. An event is created and overwriting an existing event. Do we want that? 
Yes, I know that it is tricky to pass the error condition back up to the caller chain and eventually block sending an event.

Also the error codes here are triggered by an input event, not the FCU. Who will look at the error code?

MartinDaCosta53: Of course it should!  Don't know how I messed that up.  Now fixed

MartinDaCosta53: No.  If the event table is full then we break after sending the error message so nothing should be overwritten.

MartinDaCosta53: Yes, good point about "break".  What if the event table full returned a value of EE_MAX_EVENTS +1 at line 52?  This could then be caught in the function at line 90 to stop an over-write.

SvenRosvall: Yes, but then that value needs to be handled in findOrCreateEventByEv(). And you need to stop the sendEvent() functions from sending this.

All of this would be easy if we had exceptions. Then you just return back to the main loop and continue with the next action.

MartinDaCosta53: I agree that this should never happen if the designer seta the EE_MAX_EVENTS to be large enough.  Maybe we park in the "too difficult" file for now.

src/EventProducerService.cpp
```diff
@@ -43,11 +43,12 @@ byte EventProducerService::createDefaultEvent(byte evValue)
   unsigned int nodeNum = module_config->nodeNum;
   
   byte index = module_config->findEventSpace();
-  //eventNum = (int)index + 1;
   for (eventNum = 1; eventNum <= module_config->EE_MAX_EVENTS; eventNum++)
   {
     if (module_config->findExistingEvent(nodeNum, eventNum) == module_config->EE_MAX_EVENTS)
     {
+      controller->sendCMDERR(CMDERR_TOO_MANY_EVENTS);
+      controller->sendGRSP(OPC_RQNPN, getServiceID(), CMDERR_TOO_MANY_EVENTS);
```
SvenRosvall, line 51: Oops. I actually misread this code. findExistingEvent() returns EE_MAX_EVENTS when it cannot find the event. So in this case this is all OK and this is the ID we should use.

The error condition where the event table is full can be checked after setting index on line 45. I.e. index == EE_MAX_EVENTS.

So please remove the error messages on line 50 and 51. Or add a check for index==EE_MAX_EVENTS and put these error messages there.
I guess that we should increment the error count. But we haven't added these yet for EPService.

src/EventProducerService.cpp
SvenRosvall: Agreed.

PR #102, opened 2024-03-01
Prevent compiling for DUE case

fletchjp:
Discussed with Sven. This means that the DUE case will not fail as it does not have a file EEPROM.h


src/EepromInternalStorage.h
```diff
@@ -13,6 +13,12 @@ namespace VLCB
 class EepromInternalStorage : public Storage
 {
 public:
+
+#ifdef __SAM3X8E__
+// If SAM3X8E is defined then this file shall be compilable but this class shall not be instantiatable and will not compile if used anyway.
+  DueEepromEmulationStorage() = delete;
```
SvenRosvall, line 19: Sorry, copy error here. Please change this to EepromInternalStorage.

PR #103, opened 2024-03-03
Minor changes to suppress warnings

fletchjp:
I have made two changes. One suppresses an unused parameter warning in service.h. The other one provides a default case for the outer switch statement in MinimumNodeService.cpp in function process().

### DIFF
--- a/src/MinimumNodeService.cpp
+++ b/src/MinimumNodeService.cpp
@@ -147,6 +147,10 @@ void MinimumNodeService::process(const Action *action)
       
       case ACT_MESSAGE_IN:
         handleMessage(&action->vlcbMessage);
+		break;
+
+      default:
+          break;
 
     }
   }
--- a/src/Service.h
+++ b/src/Service.h
@@ -16,7 +16,7 @@ struct Action;
 class Service
 {
 public:
-  virtual void setController(Controller *controller) {}
+  virtual void setController(Controller * /*controller*/) {}
   virtual void begin() {}
   virtual byte getServiceID() = 0;
   virtual byte getServiceVersionID() = 0;


SvenRosvall:
Thanks for this John. I have actually decided to ignore these warnings. There are many more switch-es that would need a default. There are also a lot of these unused parameters.
Martin, do you have a thought on this?

fletchjp:
Sven, I was cleaning up things in VCANSAM3X8E and these popped up. There are unused parameters in a lot of libraries. Is it possible that the switch case is more serious, with some potential for problems? One reason for cleaning these up is that it is easier to see when there are serious problems.

fletchjp:
I leave it to you as the policy maker for the main library whether you want to go down this road.

MartinDaCosta53 (commented):
I have no strong view on this although it is probably good practice to have a default so that nothing is left undefined.  As Sven has said, there are probably numerous cases where this needs to be looked at as we tidy up the design.  The important thing is that any default does not prevent other services seeing the message if it is not acted upon here.  For example, the Mode op-code is used by other services as well as MNS

SvenRosvall:
John, I reviewed the compiler warnings to see what you saw. One funny thing is that it complains about switch on action but not on any other enum types. 
Turns out this is because the variable used for actions is the enum type. When we switch on op-code that variable is of type byte which means the compiler doesn't know what values are possible. 

I have fixed the warnings for switch (action) in a separate pull request.

I also see how the "controller" variable in the Service interface comes up many times. I'll think about how to shut this up while preserving the name.

SvenRosvall:
Reviewed the parameter name for Service::setController() and decided to accept this. So will merge this one. The PR #105 adds a few more cases where the warnings John point to show up.

PR #106, opened 2024-03-12
Clarify code after review with Martin DC

SvenRosvall:
This change contains three changes that came about after a review with Martin Da Costa.  They contain clarifying comments, simplifying the lookup functions for EV1 and removing duplicate code. To see the individual changes please look at the individual commits one by one.

### DIFF
--- a/src/Configuration.cpp
+++ b/src/Configuration.cpp
@@ -255,17 +255,7 @@ void Configuration::makeEvHashTable()
 
   for (byte idx = 0; idx < EE_MAX_EVENTS; idx++)
   {
-    readEvent(idx, evarray);
-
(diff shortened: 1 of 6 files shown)


MartinDaCosta53:
Are we sure that we want to make "findExistingEventByEv(byte evnum, byte evval)" specific to EV1? It certainly makes things simpler but precludes the use of this function to search for other EV# by some future, yet to be thought of, service. Maybe we worry about that when such a need arises.

SvenRosvall:
Had a discussion with Martin. I am going to cherry-pick the two commits about clarifications and code duplication. We will leave the EV1 change for later. This might be refactored in a different way.

PR #112, opened 2024-04-04
Add SimpleCLIUserInterface to VLCB-Arduino

fletchjp:
This adds a new version of the SerialUserInterface which allows for more use of the command line. It supports commands with arguments and can be used to build a help system.

I have provided an interface to the previous SerialUserInterface commands.

The example shows how users can add extra commands. This example runs on a MEGA 1280 or MEGA 2560. It is too large for a UNO or NANO.

### DIFF
--- a/src/SimpleCLIUserInterface.cpp
+++ b/src/SimpleCLIUserInterface.cpp
@@ -17,20 +17,9 @@ namespace VLCB
 // Create CLI Object
 SimpleCLI simpleCli(CLI_COMMAND_QUEUE_SIZE); // Default of 10 commands
 
-// Callback function for exit command
-/*
-void exitHelpCallback(cmd* c) {
-    Command cmd(c); // Create wrapper object
-
-    Serial.println("Exit help");
-	// I need a way to exit the help system from here.
-}
-*/
-
 Command cmdHelp;
 Command cmdHelpWithArg;
 Command cmdSerialInterface;
-//Command cmdExit;
 
 //////////////////////////////////////////////////////////////////
 // See Modern C++ Design, Andrei Alexandrescu, 2001(!) page 116
@@ -76,7 +65,6 @@ void SimpleCLIUserInterface::serialCallback(cmd *c) {
 	// Copy the char into the instance.
 	serialChar = local_serialChar;
     Serial.print(serialChar);
-    //Serial.print(cmd.getName()); //(same as toString())
     Serial.print(" : ");
     Serial.println(cmd.toString());
 	processSerialInputImpl(serialChar); 
@@ -86,11 +74,8 @@ void local_serialCallback(cmd *c) {
     Command cmd(c); // Create wrapper object
     Serial.print("called as ");
     Serial.print(local_serialChar);
-    //Serial.print(cmd.getName()); //(same as toString())
     Serial.print(" : ");
     Serial.println(cmd.toString());
-	// How can I get access to the instance of simpleUserInterface??
-	// The answer is to use local_simpleUserInterface.
 	(simpleCLIUserInterface.*pSerialCallback)(c);
 }
 
@@ -109,15 +94,11 @@ void SimpleCLIUserInterface::setupHelp()
     cmdSerialInterface = simpleCli.addCommand("n,e,v,c,h,y,s,z,*",local_serialCallback);
 	cmdSerialInterface.setDescription("serial commands n, e, v, c, h, y, s, z and * for the module");
 	
-
-    //Serial.println(cli.toString());  // Prints all the help items.
     Serial.println("> VLCB help system started!!");
 }
 
 void SimpleCLIUserInterface::getHelp()
 {
-   //Serial << "VLCB help system available" << endl;
-	 // Ideally this should be in a separate task allowing other things to go on.
    if (Serial.available()) {
         String input = Serial.readStringUntil('\n');
 
@@ -169,8 +150,6 @@ void SimpleCLIUserInterface::getHelp()
 
 void SimpleCLIUserInterface::processSerialInput()
 {
-  //byte uev = 0;
-  //char msgstr[32], dstr[32];
 
   if (Serial.available())
   {
@@ -185,10 +164,6 @@ void SimpleCLIUserInterface::processSerialInputImpl(char c)
 {
   byte uev = 0;
   char msgstr[32], dstr[32];
-
-  //if (Serial.available())
-  //{
-  //  char c = Serial.read();
 
     switch (c)
     {
@@ -323,15 +298,11 @@ void SimpleCLIUserInterface::processSerialInputImpl(char c)
         Serial << endl;
         break;
 
-      //case 'x':
-	  //  getHelp();
-	  //break;
-
       default:
         Serial << F("> unknown command ") << c << endl;
         break;
     }
-//}
+
 }
 
 void SimpleCLIUserInterface::handleAction(const Action *action)
--- a/src/SimpleCLIUserInterface.cpp
+++ b/src/SimpleCLIUserInterface.cpp
@@ -1,0 +1,381 @@
+// Copyright (C) Sven Rosvall (sven@rosvall.ie)
+// This file is part of VLCB-Arduino project on https://github.com/SvenRosvall/VLCB-Arduino
+// Licensed under the Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License.
+// The full licence can be found at: http://creativecommons.org/licenses/by-nc-sa/4.0/
+
+#include "SimpleCLIUserInterface.h"
+#include "Controller.h"
+#include <Streaming.h>
+#include <SimpleCLI.h>
+
+extern void printConfig();
+
+
+namespace VLCB
+{
+
+// Create CLI Object
+SimpleCLI simpleCli(CLI_COMMAND_QUEUE_SIZE); // Default of 10 commands
+
+// Callback function for exit command
+/*
+void exitHelpCallback(cmd* c) {
+    Command cmd(c); // Create wrapper object
+
+    Serial.println("Exit help");
+	// I need a way to exit the help system from here.
+}
+*/
+
+Command cmdHelp;
+Command cmdHelpWithArg;
+Command cmdSerialInterface;
+//Command cmdExit;
+
+//////////////////////////////////////////////////////////////////
+// See Modern C++ Design, Andrei Alexandrescu, 2001(!) page 116
+// First declare a typedef for a pointer to what is needed.
+// Note the argument type.
+typedef void (SimpleCLIUserInterface::* TpMemFun)(cmd *);
+// Then create an object which implements tha callback
+TpMemFun pSerialCallback = &SimpleCLIUserInterface::serialCallback;
+// An instance of class SimpleUserInterace is also needed.
+// Where is it? Declared in the top level code.
+// This is the actual call to use the callback
+//     (simpleCLIUserInterface.*pSerialCallback)(c);
+// It is used in local_templateCallback(cmd *c) - see below.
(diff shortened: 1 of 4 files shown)


src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,60 @@
...
+//extern char local_serial_char;
+
+} // end namespace VLCB
+
+// Allow access to the instance in SimpleCLIUserInterface.cpp
+// Note that this is not in the VLCB namespace.
+extern VLCB::SimpleCLIUserInterface simpleCLIUserInterface;
+
+//using VLCB::cli;
+//extern VLCB::SimpleCLI cli;
```
SvenRosvall, line 58: Please remove commented code.

src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,60 @@
...
+  void getHelp();
+public:
+  void serialCallback(cmd *c);
+  void setupHelp();
+  char getFirstChar() const {return serialChar; }
+  
+};
+
+extern SimpleCLI simpleCli;
+//extern char local_serial_char;
```
SvenRosvall, line 49: Please remove commented code.

fletchjp:
Deleted commented code from SimpleCLIUserInterface.h

fletchjp:
Also delete commented code in SimpleUserInterface.cpp

src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,57 @@
+// Copyright (C) Sven Rosvall (sven@rosvall.ie)
```
SvenRosvall, line 1: Please put your name in these copyright notices.

src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,57 @@
...
+  void indicateMode(VlcbModeParams i);
+  void getHelp();
+public:
+  void serialCallback(cmd *c);
+  void setupHelp();
+  char getFirstChar() const {return serialChar; }
+  
+};
+
+extern SimpleCLI simpleCli;
```
SvenRosvall, line 48: I don't like global variables. I would make the simpleCli object a member of class SimpleCLIUserInterface. It is only used within that class anyway.

src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,57 @@
...
+  
+  void handleAction(const Action *action);
+  void processSerialInput();
+  void processSerialInputImpl(char c);
+  void indicateMode(VlcbModeParams i);
+  void getHelp();
+public:
+  void serialCallback(cmd *c);
+  void setupHelp();
+  char getFirstChar() const {return serialChar; }
```
SvenRosvall, line 44: Can you put these public methods together with the other public methods above?

src/SimpleCLIUserInterface.h
```diff
@@ -0,0 +1,57 @@
...
+  
+};
+
+extern SimpleCLI simpleCli;
+
+} // end namespace VLCB
+
+// Allow access to the instance in SimpleCLIUserInterface.cpp
+// Note that this is not in the VLCB namespace.
+extern VLCB::SimpleCLIUserInterface simpleCLIUserInterface;
```
SvenRosvall, line 54: I don't like global variables. Announcing them in header files is even worse. This simpleCLIUserInterface is only used within SimpleCLIUserInterface for the callback. You can hide this variable as a static inside SimpleCLIUserInterface.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+  getHelp();
+}
+
+// This will hold a local copy of the first char of the input.
+char local_serialChar = '0'; 
+
+void SimpleCLIUserInterface::serialCallback(cmd *c) {
+    Command cmd(c); // Create wrapper object
+    Serial.print("called as ");
+	// Copy the char into the instance.
```
SvenRosvall, line 65: Looks like you have tabs in your code. This is an example that shows why tab characters should be avoided. The indentation here and in a few other places get messed up. Please set a flag in your editor to replace tabs with spaces.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+void SimpleCLIUserInterface::process(const Action *action)
+{
+  handleAction(action);
+  
+  //This replaces processSerialInput(); 
+  getHelp();
+}
+
+// This will hold a local copy of the first char of the input.
+char local_serialChar = '0';
```
SvenRosvall, line 60: I don't like global variables. At the least make this a "static" to reduce its scope to this file. Better still is to make it a member of the class SimpleCLIUserInterface.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+	// Copy the char into the instance.
+	serialChar = local_serialChar;
+    Serial.print(serialChar);
+    Serial.print(" : ");
+    Serial.println(cmd.toString());
+	processSerialInputImpl(serialChar); 
+}
+
+void local_serialCallback(cmd *c) {
+    Command cmd(c); // Create wrapper object
```
SvenRosvall, line 74: Try to avoid using a variable name with the same name as a type. This is confusing.
It is bad that the SimpleCLI creates types with all lower case letters.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+namespace VLCB
+{
+
+// Create CLI Object
+SimpleCLI simpleCli(CLI_COMMAND_QUEUE_SIZE); // Default of 10 commands
+
+Command cmdHelp;
+Command cmdHelpWithArg;
+Command cmdSerialInterface;
```
SvenRosvall, line 22: Can you make these 4 objects members of SimpleCLIUserInterface? 
Avoid global variables.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+//////////////////////////////////////////////////////////////////
+// See Modern C++ Design, Andrei Alexandrescu, 2001(!) page 116
+// First declare a typedef for a pointer to what is needed.
+// Note the argument type.
+typedef void (SimpleCLIUserInterface::* TpMemFun)(cmd *);
+// Then create an object which implements tha callback
+TpMemFun pSerialCallback = &SimpleCLIUserInterface::serialCallback;
```
SvenRosvall, line 30: Please make pSerialCallback "static" to reduce scope to this file.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+	processSerialInputImpl(serialChar); 
+}
+
+void local_serialCallback(cmd *c) {
+    Command cmd(c); // Create wrapper object
+    Serial.print("called as ");
+    Serial.print(local_serialChar);
+    Serial.print(" : ");
+    Serial.println(cmd.toString());
+	(simpleCLIUserInterface.*pSerialCallback)(c);
```
SvenRosvall, line 79: simpleCLIUserInterface is a global variable that must have a name that matches what is defined in the user sketch. Firstly I don't like globals. Secondly you are going to run into problems where users suddenly decide to choose a shorter name and then get a cryptic message from the linker.

Instead I suggest that you have a file static pointer in this file. The SimpleCLIUserInterface constructor sets this file static pointer to "this".

SvenRosvall:
Looking at how SimpleCLIUserInterface fits in here. I actually think it shouldn't be within the VLCB-Arduino library. It has a dependency on SimpleCLI. This means that SimpleCLI library must be available to compile the VLCB-Arduino library even if SimpleCLIUserInterface is not used. 
We have a similar problem with CAN2515 which has a dependency on ACAN2515. We are planning on moving CAN2515 out to a separate library to avoid this dependency. So the same argument applies to SimpleCLIUserInterface.
I still think this is an interesting addition to the VLCB library.

src/SimpleCLIUserInterface.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+void SimpleCLIUserInterface::serialCallback(cmd *c) {
+    Command cmd(c); // Create wrapper object
+    Serial.print("called as ");
+	// Copy the char into the instance.
+	serialChar = local_serialChar;
+    Serial.print(serialChar);
+    Serial.print(" : ");
+    Serial.println(cmd.toString());
+	processSerialInputImpl(serialChar);
```
SvenRosvall, line 70: I thought the whole idea with SerialCLI was to define commands that will execute so that you don't have to have this massive processSerialInputImpl() function. I.e. define the command "n" that then prints out node information, etc.

src/SimpleCLIUserInterface.cpp
fletchjp: I think that the static version does not work - see comments in the example and previous email conversations.

src/SimpleCLIUserInterface.cpp
fletchjp: processSerialInputImpl is a callback function which does all of the commands.  Otherwise it would be necessary to have separate callbacks for each single char command. I have reused the existing code to process the char input.

src/SimpleCLIUserInterface.h
fletchjp: I have started to do that in the example files. I think we should agree who should be in what sections. If (as I now agree) this is to be a separate repository I want to agree a name for the repository as well.

src/SimpleCLIUserInterface.h
fletchjp: O.K.  I was keeping distinct the new ones added (these) as compared with those in SerialUserInterface.

src/SimpleCLIUserInterface.h
fletchjp: It is used in the client code to set up the command system (line 152). It is also used in the client callback example (line 172). I am attempting to have a system which the writer of application code can adapt as needed. I have spent some time experimenting with examples to get a workable system.

src/SimpleCLIUserInterface.cpp
fletchjp: I think that this can be resolved. The purpose of these char values is to enable the user to have access to which character was typed in a command such as "a,b" which can be called as 'a' or 'b'. This is how, for example, I have implemented the main callback to do the VLCB commands. One command can have many character inputs. Then by finding which one was used, the correct action can be taken.

src/SimpleCLIUserInterface.cpp
fletchjp: I have been following the examples which came with the library. We have already encountered the problem with "cli", In this part of the code I was attempting to get from a callback outside the main class to one inside it. I had decided that I did not want to make any changes to the third party library SimpleCLI so I had to figure out a solution. See some comments in the example code.

src/SimpleCLIUserInterface.cpp
fletchjp: I am not sure what the consequences would be for the way SimpleCLI works  They are in the VLCB namespace. I suggest you adopt the phrase "Avoid global variables where possible" rather than just lecturing me.

src/SimpleCLIUserInterface.cpp
fletchjp: I have reservations about the way commenting works other than for detail code comments. It has resulted in a very bitty discussion which has resulted in a very bitty set of replies from me. I agree that this code should be a separate library from VLCB-Arduino to avoid introducing the dependeny on SimpleCLI when it is not needed.

SvenRosvall:
Regecting this pull request as John will create this extension in a separate library. See earlier comments.

PR #116, opened 2024-08-16
Fix situations where currentMode is SETUP.

SvenRosvall:
Sometimes when I reprogram my Arduino it starts up with a flashing yellow LED. Turns out the EEPROM has been zeroed and thus Configuration->currentMode is set to SETUP, which is an illegal state for "currentMode". (Only "instantMode" can be in SETUP.)


MartinDaCosta53:
Good catch and an elegant fix.  There is always a potential problem with re-using an Arduino or any other microcontroller where the state of the EEPROM is unknown.  I often do a factory reset so as to get to a clearly defined state.

PR #117, opened 2024-08-16
Remove event services from VLCB_empty example sketch.

SvenRosvall:
Make VLCB_empty sketch smaller. It has no NVs or events so remove the NV service and the event services.


MartinDaCosta53 (approved):
I think that this is a good move as it allows users to compare with 1in1out, etc. and see how services can be added.

PR #119, opened 2024-08-19
Document example sketches.

SvenRosvall:


MartinDaCosta53:
line 6 defines
line 21 / 22 isn't quite true in that Serial GC example does not use CAN
line 29 suggest "It will demonstrate...
line 59 File extension should be .ino

SvenRosvall:
Thanks. Fixed those typos now.

MartinDaCosta53 (approved):
Happy with this.

PR #120, opened 2024-09-03
Refactorings to reduce code size

SvenRosvall:
Two refactorings that reduce code size a little.
* Change types from builtin types to the enum types defined in vlcbdefs.hpp. Some of these types are smaller than the "int" used earlier.
* Handling mode command for setting heartbeat on/off deduplicated.

### DIFF
--- a/src/CanService.h
+++ b/src/CanService.h
@@ -22,7 +22,7 @@ public:
   CanService(CanTransport * tpt) : canTransport(tpt) {}
 
   virtual void setController(Controller *cntrl) override;
-  virtual byte getServiceID() override { return SERVICE_ID_CAN; }
+  virtual VlcbServiceTypes getServiceID() override { return SERVICE_ID_CAN; }
   virtual byte getServiceVersionID() override { return 1; }
 
   virtual void process(const Action * action) override;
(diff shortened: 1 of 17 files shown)


MartinDaCosta53 (approved):
This also brings more consistency to the code eg opcode is sometimes a byte and others an int!

MartinDaCosta53 (commented):
I haven't seen the syntax in the new line 501 before.  Is this an implied if statement?

SvenRosvall:
> I haven't seen the syntax in the new line 501 before. Is this an implied if statement?

It is just a boolean (logical) expression being assigned to noHeartbeat. Looking at it now, I think I should have made it clearer with some parenthesis:
`noHeartbeat = (requestedMode == MODE_HEARTBEAT_OFF);`

It works because of the precedense rules in C++. "==" is evaluated before "=".

This is just a trick to make the code work for both ON and OFF. You can call it "implied if statement" if you want. The C++ term is just "boolean expression", i.e. just an expression that yields a boolean value.

MartinDaCosta53:
Yes, that makes it clear.  Just took a bit to get my head round it!

PR #122, opened 2024-11-02
Refactor access to Controller object in services.

SvenRosvall:
This is a refactoring to make things easier to implement diagnostics. 
Instead of each service having a pointer to the Controller, add this controller to the base Service class.
No functional changes.


MartinDaCosta53 (approved):
Looking at my VCAN2040 library, I don't think that I need to change anything as a result of this PR.

PR #123, opened 2024-11-10
Implement MNS diagnostics

SvenRosvall:
Implement some diagnostics for MNS. This is the beginning of diagnostics. More will come later.

### DIFF
--- a/src/MinimumNodeService.cpp
+++ b/src/MinimumNodeService.cpp
@@ -404,25 +404,34 @@ void MinimumNodeService::handleRequestServiceDefinitions(const VlcbMessage *msg,
 
 void MinimumNodeService::handleRequestDiagnostics(const VlcbMessage *msg, unsigned int nn)
 {
-  if (isThisNodeNumber(nn))
+  if (!isThisNodeNumber(nn))
   {
-    if (msg->len < 5)
-    {
-      controller->sendGRSP(OPC_RDGN, getServiceID(), CMDERR_INV_CMD);
-      return;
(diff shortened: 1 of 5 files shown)


SvenRosvall:
> How much extra memory does this use? Will it all fit in an Arduino Uno?

With these diagnostics we are now at Flash: 26664 bytes (86%) / Ram: 1085 bytes (52%)
This is up from 25978 bytes (84%) / 1045 bytes (51%) after the recent refactoring.
Btw, I am working on CAN Service diagnostics and this pushes us up to 27256 bytes (88%) / 1085 bytes (52%)
This will probably grow further as we add more diagnostics.

Should we make diagnostics optional?

dpharris:
I think it demonstrates that the 328 is just barely adequate. The same
holds for LCC. Unfortunately it is such an attractive platform. There are
newer versions, such as the Nano Every which uses the 4809, more flash and
RAM, but less EEPROM. Time to move to the Pico, ESP33, STM32, etc.

That said, making diags optional for 8 bit processors might be necessary to
allow them to be used implement simple modules.

Getting close to the maximum sizes makes the modules more fragile as the
stack and heap may collide unpredictably.

David

MartinDaCosta53:
I think that we need to make diagnostics optional.  I don't believe that most users will ever use the diagnostics.  It's OK on a Pico, which is what I'm now using most of the time (so cheap, why wouldn't you?) or a Mega but a Nano and Uno will have little room for an application!

PR #125, opened 2024-11-17
Update MinimumNodeService.cpp

MartinDaCosta53:
To align MNS with spec changes.

### DIFF
--- a/src/LEDUserInterface.h
+++ b/src/LEDUserInterface.h
@@ -13,7 +13,7 @@
 namespace VLCB
 {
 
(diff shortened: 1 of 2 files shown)


SvenRosvall:
I wonder if we shall also change the 6 seconds required for the push button to change mode to 4 seconds?
Otherwise looks good.

MartinDaCosta53:
Yes, we should.  Can I add a change to LEDUserInterface.h to this PR or should I merge and open another?

MartinDaCosta53:
OK, added the time change.

MartinDaCosta53:
All done

PR #126, opened 2024-11-19
Make diagnostics optional.

SvenRosvall:
This saves a lot of memory if diagnostics is not required.

This adds 80 bytes of overhead for splitting out the diagnostics into child classes when diagnostics is used.
But saves 1200 bytes when diagnostics is not used.


MartinDaCosta53 (approved):
This seems to be a good way of continuing to allow a VLCB to run on the smaller Arduinos and, perhaps, other small processors.  I'm inclined not to offer the no diagnostics option on VCAN2040 as it is only appropriate for a large capacity processor, but I'll listen to contrary arguement.

PR #127, opened 2024-11-21
Update MinimumNodeService.cpp

MartinDaCosta53:
Align ACT_RENOGIATE actions with spec.

### DIFF
--- a/src/MinimumNodeService.cpp
+++ b/src/MinimumNodeService.cpp
@@ -23,7 +23,7 @@ void MinimumNodeService::begin()
 //
 /// initiate the transition from Uninitialised to Normal mode
 //
-void MinimumNodeService::initSetup()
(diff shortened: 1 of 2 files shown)


src/MinimumNodeService.cpp
```diff
@@ -139,10 +139,23 @@ void MinimumNodeService::process(const Action *action)
         break;
       
       case ACT_RENEGOTIATE:
-        switch (controller->getModuleConfig()->currentMode)
+        switch (instantMode)
         {
         case MODE_UNINITIALISED:
           break;
+          
+        case MODE_SETUP:
```
SvenRosvall, line 147: Please add a comment that this action cancels the setup and reverts to previous mode.

MartinDaCosta53: Done and other change we discussed.

PR #131, opened 2024-11-25
Small updates to reduce code size slightly.

SvenRosvall:
Found a few small bits of duplicate code.

### DIFF
--- a/src/MinimumNodeService.cpp
+++ b/src/MinimumNodeService.cpp
@@ -27,14 +27,19 @@ void MinimumNodeService::initSetupFromUninitialised()
 {
   // DEBUG_SERIAL << F("> initiating Normal negotation") << endl;
 
+  // enumerate the CAN bus to allocate a free CAN ID
+  Action action = {ACT_START_CAN_ENUMERATION, false };
+  controller->putAction(action);
+
+  initSetupCommon();
+}
+
+void MinimumNodeService::initSetupCommon()
+{
   instantMode = MODE_SETUP;
   controller->indicateMode(MODE_SETUP);
 
   timeOutTimer = millis();
-  
-  // enumerate the CAN bus to allocate a free CAN ID
-  Action action = {ACT_START_CAN_ENUMERATION, false };
-  controller->putAction(action);
 
   // send RQNN message with current NN, which may be zero if a virgin/Uninitialised node
   controller->sendMessageWithNN(OPC_RQNN);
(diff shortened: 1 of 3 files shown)


src/MinimumNodeService.cpp
```diff
@@ -57,20 +62,18 @@ void MinimumNodeService::setNormal(unsigned int nn)
...
-  requestingNewNN = false;
-  instantMode = MODE_UNINITIALISED;
   if (controller->getModuleConfig()->nodeNum != 0)
   {
     controller->sendMessageWithNN(OPC_NNREL);  // release node number first
   }
+  requestingNewNN = false;
+  instantMode = MODE_UNINITIALISED;
   controller->getModuleConfig()->setModuleUninitializedMode();
-  controller->getModuleConfig()->setCANID(0);
```
SvenRosvall, line 67: MNS shouldn't really know about CANID. It is a configuration so belongs there. Having this code in configuration means that the code for dereferencing ( operator "->" ) is not needed.

src/MinimumNodeService.cpp
```diff
@@ -27,14 +27,19 @@ void MinimumNodeService::initSetupFromUninitialised()
...
   // DEBUG_SERIAL << F("> initiating Normal negotation") << endl;
 
+  // enumerate the CAN bus to allocate a free CAN ID
+  Action action = {ACT_START_CAN_ENUMERATION, false };
+  controller->putAction(action);
+
+  initSetupCommon();
+}
+
+void MinimumNodeService::initSetupCommon()
```
SvenRosvall, line 37: The bits here are also used by initSetupFromNormal() below.

src/MinimumNodeService.cpp
MartinDaCosta53: I agree.  CANID is also transport specific and so whilst it needs to be stored using Configuration, it may not be called that by other transports.  It might not even exist!  A thought for going forward as it is not an issue, yet.

src/MinimumNodeService.cpp
MartinDaCosta53, line 1: I'm not sure that it is a good idea to leave instantMode = MODE_UNINITIALISED when we set uninitialised, which is what will happen if you delete line 61.  The effect of this will to cause a short button press to initiate a call to initSetupFromNormal() at line 174, which is not what we want.

src/MinimumNodeService.cpp
MartinDaCosta53: I agree.  Good catch!

src/MinimumNodeService.cpp
SvenRosvall: Please note that I re-arranged the lines in setUninitialised(). instantMode is set on line 70. 
I re-arranged these lines so that setUninitialised() and setNormal() would look more similar. And maybe this would be another candidate for deduplication. Well, it wasn't.

MartinDaCosta53: So you did!  Sorry, I missed that it had been moved and not deleted.

MartinDaCosta53 (approved):

PR #133, opened 2024-11-27
Omit NNREL when not needed.

SvenRosvall:


MartinDaCosta53 (commented):
In reset to manufacturer's defaults, if you don't send NNREL then the FCU will not know that the module has been set to Uninitialised.  That is why I put that line in.

I had been looking at removing requestingNewNN, as well.  Saves me doing that change!

PR #137, opened 2024-12-04
Minor code improvements

SvenRosvall:
These are some improvements I have accumulated over some time.

- "const" methods. Marking a method "const" tells the compiler (and code reader) that this method will not change the object it belongs to.
- Remove declarations of class Configuration. These declarations were needed when the service classes owned a reference to a Controller object, and for efficiency also a Configuration object. We moved this to the base class Service some time ago. This declaration should have been removed at that time.
- Removed an unused method declaration.
- Fix a constant in setEventAck(). 

The last item is the only change that affects the behaviour of the code. All others are just cleanups.

### DIFF
--- a/src/ArrayHolder.h
+++ b/src/ArrayHolder.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <stddef.h>
 #include "initializer_list.h"
 
 namespace VLCB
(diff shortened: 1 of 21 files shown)


MartinDaCosta53 (approved):
What does making methods const do for us?

PR #139, opened 2024-12-12
Rearrange actions for early function returns.

SvenRosvall:
Please review this with "Hide whitespaces" checked to minimize the amount of visible changes.


MartinDaCosta53 (approved):
This all makes sense.  I'm sure that we have done a similar exercise before.

PR #140, opened 2024-12-14
Update EventTeachingService.cpp

MartinDaCosta53:
Prevent duplication of producer events

### DIFF
--- a/src/EventTeachingService.cpp
+++ b/src/EventTeachingService.cpp
@@ -415,23 +415,23 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
     controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_EV_IDX);
     return;
   }
+  
+  byte index = module_config->findExistingEvent(nn, en);
+    
   // Is this a produced event that we know about?
   // Search the events table by evnum = 1 for a value match with evval.
   if ((evnum == 1) && (evval > 0))
   {
-    byte index = module_config->findExistingEventByEv(evnum, evval);
-    if (index < module_config->EE_MAX_EVENTS)
+    byte indexEV1 = module_config->findExistingEventByEv(evnum, evval);
+    if (indexEV1 < module_config->EE_MAX_EVENTS)
     {
       // Ensure that we won't create multiple events for the evval
       // Search for this NN, EN
-      byte indexNE = module_config->findExistingEvent(nn, en);
-      if (indexNE != index)
+      byte index = module_config->findExistingEvent(nn, en);
+      if (index != index)
       {
         // respond with error
-        controller->sendCMDERR(CMDERR_INV_EV_VALUE);  // Deprecated in favour of GRSP_OK
-        // DEBUG_SERIAL <<F("ets> WRACK sent") << endl;
-
-        // Note that the op-code spec only lists WRACK as successful response.
+        controller->sendCMDERR(CMDERR_INV_EV_VALUE);
         controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_EV_VALUE);
         return;        
       }
@@ -440,8 +440,6 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
       
   // search for this NN, EN as we may just be adding an EV to an existing learned event 
   //DEBUG_SERIAL << F("ets> searching for existing event to update") << endl;
-  byte index = module_config->findExistingEvent(nn, en);
-
   // not found - it's a new event
   if (index >= module_config->EE_MAX_EVENTS)
   {
--- a/src/EventTeachingService.cpp
+++ b/src/EventTeachingService.cpp
@@ -417,7 +417,7 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
   }
   
   byte index = module_config->findExistingEvent(nn, en);
-    
+  
   // Is this a produced event that we know about?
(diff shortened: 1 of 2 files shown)


SvenRosvall:
Hmm, this looks broken. Yes you have added a condition to check for duplicate events. But the old code for adding an event is gone. I see no call to writeEvent() in the new code.

src/EventTeachingService.cpp
```diff
@@ -422,19 +422,19 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
...
-      controller->sendGRSP(OPC_EVLRN, getServiceID(), GRSP_OK);
-      ++diagEventsTaught;
-      return;
+      // Ensure that we won't create multiple events for the evval
+      // Search for this NN, EN
+      byte indexNE = module_config->findExistingEvent(nn, en);
+      if (indexNE != index)
+      {
+        // respond with WRACK
+        controller->sendCMDERR(CMDERR_INV_EV_VALUE);  // Deprecated in favour of GRSP_OK
```
SvenRosvall, line 431: The comments on these three lines talk about WRACK. I assume this is carried over from the old code.

MartinDaCosta53:
We don't write anything here now.  This is just a trap to stop duplicates.  The event is now only written down at line 468

src/EventTeachingService.cpp
MartinDaCosta53: Oops! Sorry.  Now corrected

SvenRosvall: Two more comments. 
Line 431: // Deprecated in favour of GRSP_OK
You can remove this comment.
Line 432: // DEBUG_SERIAL <<F("ets> WRACK sent") << endl;
Not accurate. Don't think you need this debug output anyway.

SvenRosvall: Do you need both CMDERR and GRSP?

SvenRosvall:
> We don't write anything here now. This is just a trap to stop duplicates. The event is now only written down at line 468

OK, I understand how the old code worked now. It was just updating the NN/EN and then returning.

src/EventTeachingService.cpp
```diff
@@ -422,19 +422,19 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
...
-      controller->sendWRACK();  // Deprecated in favour of GRSP_OK
-      // DEBUG_SERIAL <<F("ets> WRACK sent") << endl;
-
-      // Note that the op-code spec only lists WRACK as successful response.
-      controller->sendGRSP(OPC_EVLRN, getServiceID(), GRSP_OK);
-      ++diagEventsTaught;
-      return;
+      // Ensure that we won't create multiple events for the evval
+      // Search for this NN, EN
+      byte indexNE = module_config->findExistingEvent(nn, en);
```
SvenRosvall, line 427: A suggestion for reducing duplication:
This indexNE is the same as the index on line 443. Can you move that line up before line 420. Then rename the index on line 422 to (say) indexEV1? 
This way you don't need to call findExistingEvent() twice.

src/EventTeachingService.cpp
MartinDaCosta53: No, but having both makes for consistency with what is already done elsewhere.

src/EventTeachingService.cpp
```diff
@@ -422,19 +422,19 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
...
+      if (indexNE != index)
```
SvenRosvall, line 428: I checked out this PR to run the unit tests. There is a failure here when updating NN/EN for an existing event. It fails where the new NN/EN is not found. This means that the indexNE here is set to  EE_MAX_EVENTS. This is different from the index above (that I want to rename to indexEV1) which is 0, i.e. the first event. 
Suggest changing this condition to:
`   if (indexNE != module_config->EE_MAX_EVENTS || indexNE != index)`
However this is not enough for this case. Now we come to line 446 where we check the index from findExistingEvent() (with the new NN/EN). There is no such event and we search for a free slot with findEventSpace(). This succeeds and we create a new event at index==1. This means that first of all we don't update the event we intended. Secondly we have a second event with the same EV1 value but with a different NN/EN.
So the logic here seems to be a lot more complex.

src/EventTeachingService.cpp
MartinDaCosta53: Good catch!

src/EventTeachingService.cpp
MartinDaCosta53: You're right about the condition, but it no longer works properly on my test rig.  It did this morning.  I need to unwind the changes and see where it breaks.  May take until tomorrow.

src/EventTeachingService.cpp
```diff
@@ -415,33 +415,43 @@ void EventTeachingService::handleLearnEvent(const VlcbMessage *msg, unsigned int
...
+      else
+      {
+        if (index != indexEV1)
+        {
+          // respond with error. Producer EV value assigned to another NN/EN
+          controller->sendCMDERR(CMDERR_INV_EV_VALUE);
+          controller->sendGRSP(OPC_EVLRN, getServiceID(), CMDERR_INV_EV_VALUE);
+          return;
+        }
+        //else valid learn so proceed
```
SvenRosvall, line 447: Here the two indexes are equal. This means that the NNEN is the same and the value of EV1 is the same as the stored event. Do we need to proceed here? 
We chatted briefly about this and we could return here as the event will be updated with the same value for EV1.  No need to write this.
No functional change, just a small optimization.

PR #142, opened 2024-12-18
Rearrange code for early returns.

SvenRosvall:
Found a few more occurrences.


MartinDaCosta53:
Looking at these, I don't see how you get an earlier return in the processor, but it does get the return in earlier in the code.  It also reduces the level of nesting, which is always good, I think, and gives a consistent approach.

PR #144, opened 2024-12-27
Limit interrupt blocking by EEPROM commit to minimum possible time

MartinDaCosta53:


src/Storage.h
```diff
@@ -21,6 +21,7 @@ class Storage
   virtual byte readBytes(unsigned int eeaddress, byte nbytes, byte dest[]) = 0;
   virtual void writeBytes(unsigned int eeaddress, const byte src[], byte numbytes) = 0;
   virtual void reset() = 0;
+  virtual void commitWriteEEPROM() = 0;
```
SvenRosvall, line 24: This line says that all child classes of Storage MUST override commitWriteEEPROM(). 
I am surprised you could compile as the other child classes are compiled even if they are not used in your project.

src/EepromInternalStorage.cpp
```diff
@@ -75,10 +75,15 @@ void EepromInternalStorage::write(unsigned int eeaddress, byte data)
 //
 void EepromInternalStorage::writeBytes(unsigned int eeaddress, const byte src[], byte numbytes)
 {
+  #ifdef __SAM3X8E__
+  // Shouldn't use this for the Due.
+  #else
   for (byte i = 0; i < numbytes; i++)
   {
-    setChipEEPROMVal(eeaddress + i, src[i]);
+    EEPROM.write(eeaddress + i, src[i]);
```
SvenRosvall, line 83: You can change this back to use setChipEEPROMVal() now that it doesn't commit. Then you can remove the #ifdef for __SAM3X8E__. It is enough to have that #ifdef in the low level functions.

SvenRosvall:
That is a small and effective fix. Just added some comments to make it even smaller.

src/Storage.h
SvenRosvall: Oh, sorry, I am wrong. I remember now how to do default implementation of virtual functions. We put the default implementation in the interface class (Storage) and then there is no need to change any child classes. 
So Storage.h should include:
`virtual void commitWriteEEPROM() {}`
Note the {} at the end. And no "override" here. Then you can revert the changes in the child classes.

PR #146, opened 2025-01-17
Determine maximum use of EEPROM and make available to user application

MartinDaCosta53:


src/Configuration.h
```diff
@@ -82,6 +84,7 @@ class Configuration
   byte EE_NUM_EVS = 0;
   byte EE_PRODUCED_EVENTS = 0;
   byte EE_BYTES_PER_EVENT; // Value calculated in begin()
+  unsigned int EE_VLCB_END; // Value calculated in begin()
```
SvenRosvall, line 87: I am trying to come up with a better name. "VLCB" is not describing this well. How about EE_CONFIG_END or EE_MODULE_END or EE_USER_START ?

src/Configuration.h
```diff
@@ -46,6 +46,8 @@ class Configuration
   byte findExistingEvent(unsigned int nn, unsigned int en) const;
   byte findEventSpace() const;
   byte findExistingEventByEv(byte evnum, byte evval) const;
+  
+  unsigned int getFreeEEbase();
```
SvenRosvall, line 50: Do we need a getter function? We don't have one for any other EE_ variable.

src/Configuration.h
SvenRosvall: Or EE_FREE_BASE to match the getter function?

SvenRosvall:
If we have user defined EE-prom storage space, then that should be taken account to in EepromInternalStorage. See line 20 where the EE-prom size is calculated. Yes, this is only for the ESP processors but I think this need to be considered.

src/Configuration.h
MartinDaCosta53: No, we don't need the getter function.  It compiles fine without it.

MartinDaCosta53:
Ok. How about this latest change to the PR?

src/Configuration.h
```diff
@@ -82,6 +82,8 @@ class Configuration
   byte EE_NUM_EVS = 0;
   byte EE_PRODUCED_EVENTS = 0;
   byte EE_BYTES_PER_EVENT; // Value calculated in begin()
+  unsigned int EE_FREE_BASE; // Value calculated in begin()
+  unsigned int EE_USER_BYTES = 0; // Specified by user in setup for ESP processors.
```
SvenRosvall, line 86: Oh, this one needs documentation so that users don't miss this.

SvenRosvall: Looked at PersistentStorage.md and it needs a lot of work. It doesn't take into account that we now provide calculated defaults.
Yes, a How it Works would be nice.

Issue #152, opened 2025-03-25
FCU Compatibility mode


SvenRosvall:
Introduce the FCU compatibility mode.
In this mode requesting param#0, NV#0, EV#0 does not return all the values, only the number of entries.

SvenRosvall:
Done. See pull request #156

PR #156, opened 2025-06-25
Add FCU compatibility mode

MartinDaCosta53:


src/EventTeachingService.cpp
```diff
@@ -344,10 +344,13 @@ void EventTeachingService::handleReadEventVariable(const VlcbMessage *msg, unsig
   {
     // Return number of EVs. This may be dynamic in other implementations.
     controller->sendMessageWithNN(OPC_NEVAL, eventIndex, evnum, module_config->EE_NUM_EVS);
-    for (byte i = 1; i <= module_config->EE_NUM_EVS; i++)
+    if (!controller->getModuleConfig()->fcuCompatible)
```
SvenRosvall, line 347: You can replace "controller->getModuleConfig()" with "module_config".

src/EventTeachingService.cpp
```diff
@@ -632,11 +635,14 @@ void EventTeachingService::handleRequestEventVariable(const VlcbMessage *msg, un
     response.data[5] = 0;
     response.data[6] = module_config->EE_NUM_EVS;
     controller->sendMessage(&response);
-    for (byte i = 1; i <= module_config->EE_NUM_EVS; i++)
+    if (!controller->getModuleConfig()->fcuCompatible)
```
SvenRosvall, line 638: You can replace "controller->getModuleConfig()" with "module_config".

src/NodeVariableService.cpp
```diff
@@ -70,10 +70,12 @@ void NodeVariableService::handleReadNV(const VlcbMessage *msg, unsigned int nn)
   if (nvindex == 0)
   {
     controller->sendMessageWithNN(OPC_NVANS, nvindex, module_config->EE_NUM_NVS);
-
-    for (int i = 1; i <= module_config->EE_NUM_NVS; ++i)
+    if (!controller->getModuleConfig()->fcuCompatible)
```
SvenRosvall, line 73: You can replace "controller->getModuleConfig()" with "module_config".

src/MinimumNodeService.h
```diff
@@ -45,6 +45,7 @@ class MinimumNodeService : public Service
   byte heartbeatSequence = 0;
   bool noHeartbeat = false;
   unsigned int heartRate = 5000;
+  bool notFcuCompatible = false;  //Compatible is default
```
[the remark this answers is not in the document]
MartinDaCosta53: Yes, I thought long and hard about this and tried a couple of approaches.  I was also guided by what we did with noHeartbeat.  The issue really was the need to be FCU compatible by default and then meant that I wanted the flag to be true when FCU compatibility was switched off.  In the end I opted for the method used, which does have a similarity in approach to heartbeat.

src/EventTeachingService.cpp
MartinDaCosta53: Of course!  I'm not sure why I did that.  I'll change it

src/NodeVariableService.cpp
MartinDaCosta53: Of course! I'm not sure why I did that. I'll change it

SvenRosvall:
See also #152

PR #157, opened 2025-08-12
Initialise MinimumNodeService::notFcuCompatible correctly.

SvenRosvall:
Noticed this when running unit tests.
MinimumNodeService::notFcuCompatible shall be in sync with Configuration::fcuCompatible at all times. Fix here ensures they are in sync even at startup.


MartinDaCosta53:
Rather obvious when it's pointed out!  Well caught.

PR #158, opened 2025-08-28
Simplify setup in user sketches

SvenRosvall:
Lots of changes.

### DIFF
--- a/Arduino/Arduino.h
+++ b/Arduino/Arduino.h
@@ -22,7 +22,7 @@ byte lowByte(unsigned int);
 #define bitToggle(value, bit) ((value) ^= (1UL << (bit)))
 #define bitWrite(value, bit, bitvalue) ((bitvalue) ? bitSet(value, bit) : bitClear(value, bit))
 void pinMode(int, PinMode);
-void digitalWrite(int, int);
+void digitalWrite(int, PinState);
 byte digitalRead(int);
 
 struct Serial_T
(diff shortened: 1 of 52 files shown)


MartinDaCosta53 (commented):
This seems to embed CAN as the transport for VLCB.  What happens for a wireless transport? Will it need a new VLCB?

MartinDaCosta53 (commented):
In ServiceFactory, why do we #include CanService.h as well as CanServiceWithDiagnostics.h?

SvenRosvall:
> In ServiceFactory, why do we #include CanService.h as well as CanServiceWithDiagnostics.h?

The factory might create either of these services so it has to know what both classes are.

But you are right, you only need CanServiceWithDiagnostics.h as it includes CanService.h. This is a matter of style. I.e. you can have one #include for each class you use. (My style here is to have one header file for each class.) 
I could clean up headers that include other header files. But then whenever something changes then I may have to add a #include for something I use but is not in any other header file anymore. 

You will probably find lots of such redundant #include lines elsewhere in the project. Not sure I actually want to clean up these.

MartinDaCosta53:
> > This seems to embed CAN as the transport for VLCB. What happens for a wireless transport? Will it need a new VLCB?
> 
> I envisage that we could just replace the CAN service with a WIFI service. But I don't want to guess how that should be done until we have figured out how VLCB would work over WIFI. Then we'll do all the required refactoring.

OK

MartinDaCosta53:
Since an existing sketch will not include VLCB.h, will it still run with these updated libraries?

SvenRosvall:
> Since an existing sketch will not include VLCB.h, will it still run with these updated libraries?

Yes. The old header files still exist in the project. The VLCB.h is just added on top of existing header files for convenience as "include-all" header file.

PR #185, opened 2026-02-26
Work indication shall flash the yellow LED in unitialised mode.

SvenRosvall:
In unitialised mode the mode is indicated with the green LED. Flashing this LED for activity leaves the LED off after the flash.
This change is in accordance with the VLCB Technical introduction document.


MartinDaCosta53 (approved):
Looks good.  Another bug bites the dust!

PR #186, opened 2026-03-07
Clear hash table after manufacturer reset.

SvenRosvall:
After manufacturer reset the events on the node are overwritten. But MMC shows the events still with NN:EN as FFFF:FFFF. This is because the EEPROM is cleared but the event hash table is not updated. The list of events returned by the node is based on the entries in the event hash table.


MartinDaCosta53 (approved):
I see that the function to do this was always there but nothing ever called it.

PR #188, opened 2026-03-27
Button press for changing mode shall fix invalid mode value.

SvenRosvall:
This is to fix invalid mode as discussed on the MERG forum.


MartinDaCosta53 (approved):
Agreed.

PR #189, opened 2026-03-27
Change RP2040 reboot method as per CBUSconfig library.

SvenRosvall:
Duncan Greenwood made this change to CBUSconfig. Looks like something that should be here too.


MartinDaCosta53 (approved):
A much better way than using the watchdog timer.

PR #192, opened 2026-05-07
Make canp in CAN2515 private.

SvenRosvall:
Made this small change so that "canp" can be private in the class CAN2515 and thus invisible to the outside world.
Noticed that this was included in Doxygen generated files which it shouldn't. 
This is a general change so making this change on a separate branch. Will merge to the UseDoxygen branch later.

### DIFF
--- a/src/CAN2515.h
+++ b/src/CAN2515.h
@@ -72,7 +72,7 @@ public:
   virtual unsigned int errorStatus() override { return canp->errorFlagRegister(); }
 
 private:
-  ACAN2515 *canp;   // pointer to CAN object so user code can access its members
+  ACAN2515 *canp;   // pointer to CAN object
   unsigned int _numMsgsSent, _numMsgsRcvd;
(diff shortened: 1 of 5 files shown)


src/CAN2515.h
MartinDaCosta53, line 1: Should the comment at line 75 now be shortened to just read "//pointer to CAN object" Since the user can no longer access its members directly.

MartinDaCosta53: Other than that, I think that there is another benefit in that this change simplifies user access to diagnostics.

MartinDaCosta53 (commented):
This looks good and simplifies the way diagnostics are accessed by the user.  Should the comment at the new line 75 in

src/CAN2515.h
SvenRosvall: Yes, fixed the comment.

MartinDaCosta53 (approved):
Looks good to me

PR #198, opened 2026-05-19
Make some helper functions static so they don't show up in generated documentation.

SvenRosvall:
I noticed that some local helper functions showed up in the function list for the VLCB namespace. They should not be shown here just like class private methods are hidden.


MartinDaCosta53 (approved):
I thought long and hard about this but in the end concluded that this is the sort of reason why you would declare a function as static.  I assume that there aren't any consequential changes for use from outside theses classes.
I see that a fair chunk of code has moved from FlashStorage.h to .cpp.  Is this to facilitate the documentation or tidying up?

SvenRosvall:
Yes the "static" keyword makes the declared thing local to the file it is declared in. It cannot be used from other files. I did check that these declared things were not used elsewhere.

Yes, I moved some declarations from the FlashStorage.h to the .cpp file as those declarations were only used in that .cpp file. No need to have them in the header file. The purpose of header files is to share declarations with other .cpp files. ;)

PR #199, opened 2026-05-20
Minor fixes

SvenRosvall:
A few small fixes that have been lingering on my computer for while.

* Update CAN1IN1OUT MDF to match ModuleDescriptor repo.
* Add note about external implementations of CanTransport.
* Make CircularBuffer methods const. These changes were recommended by my IDE. 
* Rerun Doxygen after these changes and previous changes.


src/CircularBuffer.h
```diff
@@ -86,7 +85,7 @@ void CircularBuffer<E>::put(const E &msg)
   uint8_t size = bufUse();
   if (size > hwm)
   {
-    // Tracks high water mark (hwm)
+    // Tracks high watermark (hwm)
```
MartinDaCosta53: No.  This was correct before.  A watermark is something embedded to show authenticity getting its name from the marks put into paper during manufacture.  A High Water Mark is the mark left by the highest point that the water reaches.

MartinDaCosta53 (approved):
Ive made a comment at line 88 of the Circular Buffer.  Not come across const being applied to functiosn before.  Something else that I've learnt.  Apart from my comment, I'm happy with this.

dpharris:
I would hyphenate it: high-water mark.  But sm ok with high water mark.

David

PR #203, opened 2026-05-31
Split process function

SvenRosvall:
This change is inspired by Ian's PIC library. 
The process() function in Service classes are called regularly with or without an action. Split this so that there is a process() function for background tasks and a processAction() that is called when there is an action available.
It looks like more code but the binary is actually smaller as the process() function does not need to check if an action was provided or not.

Martin, this change means that you have to update your own service for your CANCMDDC project.

This change may conflict with other recent changes so please let me do the merge.


MartinDaCosta53 (approved):
A logical move forward.

PR #204, opened 2026-06-05
Pass required configurations size to persistent storage classes.

SvenRosvall:
This is to solve compilation errors for ESP32.


src/Configuration.cpp
MartinDaCosta53, line 1: If the calculation of EE_FREE_BASE at line 76 was moved to line 62, line 63 argument could become EE_FREE_BASE + EE_USER_BASE

SvenRosvall: Great point. Not only does it save precious program memory, but setting the EE_xxx variables should be next to each other.

MartinDaCosta53 (approved):

PR #206, opened 2026-06-08
Remove unused code for Diagnostic 0.

SvenRosvall:
The reportDiagnostic() function will never be called with diagnosticCode set to zero. MNS will handle this case separately and call reportAllDiagnostics() instead.

Discovered this while working on TimedResponse. Making a separate pull request for this so that it is easier to review the changes for TimedResponse later.

### DIFF
--- a/src/MinimumNodeServiceWithDiagnostics.cpp
+++ b/src/MinimumNodeServiceWithDiagnostics.cpp
@@ -91,9 +91,6 @@ void MinimumNodeServiceWithDiagnostics::reportDiagnostics(byte serviceIndex, byt
   unsigned int diagnosticsValue;
   switch (diagnosticsCode)
   {
-    case 0x00:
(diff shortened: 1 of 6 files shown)


src/MinimumNodeServiceWithDiagnostics.cpp
MartinDaCosta53, line 1: Would it be better to change lines 52 - 86 into a function that is called by case 0x00?

SvenRosvall: Yes, this function is pretty long. It would help readability.
This part of the code is going to be reworked for TimedResponse. So not going to clean up this mess right now.

MartinDaCosta53: OK

MartinDaCosta53 (approved):

PR #207, opened 2026-06-08
CircularBuffer size is a template param instead of dynamic memory.

SvenRosvall:
This change sets the buffer size at compile time, instead of when the buffer is constructed at run time. This means that the memory used for the buffer is reported at compile time. 

In the old code the memory was allocated at runtime when the object was created. This had the risk that memory could run out at runtime and cause the program to crash.

The memory use reported by freeSRAM() is very similar in both versions so there is no big change. Actually saving 8 bytes of memory which I believe come from the management of dynamic memory.

I hope you don't mind the syntax change where the buffer size is provided when the buffer is defined and being part of the type.

### DIFF
--- a/src/CircularBuffer.h
+++ b/src/CircularBuffer.h
@@ -10,11 +10,11 @@
 namespace VLCB
 {
 
-template <typename E>
+template <typename E, int bufferCapacity = 4>
 class CircularBuffer
 {
 public:
-  explicit CircularBuffer(uint8_t bufferCapacity = 4);
+  explicit CircularBuffer();
   ~CircularBuffer();
 
   bool available() const;
@@ -25,10 +25,10 @@ public:
 
   // Diagnostic metrics access
   uint8_t bufUse() const;
-  unsigned int getNumberOfPuts() const;
-  unsigned int getNumberOfGets() const;
-  unsigned int getOverflows() const;
-  unsigned int getHighWaterMark() const;   // High Water Mark
+  unsigned int getNumberOfPuts() const { return numPuts; }
+  unsigned int getNumberOfGets() const { return numGets; }
+  unsigned int getOverflows() const { return numOverflows; }
+  unsigned int getHighWaterMark() const { return hwm; }
 
 private:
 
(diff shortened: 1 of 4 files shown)


src/CircularBuffer.h
```diff
@@ -10,11 +10,11 @@
 namespace VLCB
 {
 
-template <typename E>
+template <typename E, int bufferCapacity = 4>
```
MartinDaCosta53, line 13: Why does bufferCapacity change from a byte to a signed integer?

src/CircularBuffer.h
```diff
@@ -25,10 +25,10 @@ class CircularBuffer
 
   // Diagnostic metrics access
   uint8_t bufUse() const;
-  unsigned int getNumberOfPuts() const;
-  unsigned int getNumberOfGets() const;
-  unsigned int getOverflows() const;
-  unsigned int getHighWaterMark() const;   // High Water Mark
+  unsigned int getNumberOfPuts() const { return numPuts; }
```
MartinDaCosta53, line 28: I like this - much more straight forward

src/Controller.h
```diff
@@ -19,6 +19,15 @@
...
 namespace VLCB
 {
+
+// TODO: This is a sign something is wrong. Need to be this big to handle 
+// large number of generated messages such as when asking for node parameters.
+// Each entry uses 10 bytes so a size of 30 uses 300 bytes.
+// This will get worse for for example queries for all events.
+// Need a mechanism like TimedResponse to create messages slowly so they can
+// be sent through the transport without requiring this buffering. 
+const int ACTION_QUEUE_SIZE = 30;
```
MartinDaCosta53, line 29: I assume that this will go completely with the introduction of timed response.

src/CircularBuffer.h
SvenRosvall: "byte" is the preferred numeric type on an 8 bit machine. This is why the constructor had a capacity type of byte.
Here we are using capacity as a template parameter. The specific type doesn't matter here as it is not used at runtime. You can see this as a constant that is used to define the size of the array used as a buffer on line 46.

src/Controller.h
SvenRosvall: Yes.
Or rather, the size will become the default, which is 4 elements.

src/CircularBuffer.h
MartinDaCosta53: OK.  I'd have been less surprised if you had used uint16_t

PR #208, opened 2026-06-09
Add CI workflows

brocci:
### Summary
This PR introduces a split CI structure.

It includes:
- Host CI for CMake build and test execution
- Arduino CI for compiling all example sketches with arduino-cli
- Docs CI for generating Sketch and Library Doxygen output and uploading artifacts

### Why
The goal is to separate responsibilities clearly:
- Host logic verification via CMake tests
- Firmware build verification via arduino-cli
- Documentation generation via Doxygen

Compared to manual runs in developers' environments, this adds:
- Repeatable checks on a clean runner
- Earlier regression detection
- Shared visibility of results
- Less manual verification during review

> Note: This is an addition to the current manual workflows in developers' environments, providing automated CI checks alongside existing local validation.

#### CI and docs
- Added host test workflow
- Added Arduino build workflow
- Added docs workflow with Doxygen generation and artifact upload

### Validation
Verified on fork Actions:
- Host CMake workflow passes and executes testAll successfully
- Arduino workflow compiles all examples successfully
- Docs workflow generates and uploads both documentation artifacts successfully

### Related PR
SerialGC follow-up has been split into a separate PR:
- https://github.com/SvenRosvall/VLCB-Arduino/pull/209

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,46 @@
+name: Docs
+
+on:
+  pull_request:
+    paths:
+      - docs/**
+      - src/**
+      - Doxygen.*.conf
+      - .github/workflows/docs.yml
+  push:
+    paths:
+      - docs/**
+      - src/**
+      - Doxygen.*.conf
+      - .github/workflows/docs.yml
+  workflow_dispatch:
+
+jobs:
+  generate-doxygen:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Install documentation tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y doxygen graphviz
+
+      - name: Generate API documentation
+        run: |
+          doxygen Doxygen.Sketch.conf
+          doxygen Doxygen.Library.conf
+
+      - name: Upload sketch docs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: html-sketch
+          path: html.sketch
+
+      - name: Upload library docs artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: html-library
+          path: html.library
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,8 @@
 name: Docs
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+  
 on:
   pull_request:
     paths:
--- a/.github/workflows/ci-arduino-build.yml
+++ b/.github/workflows/ci-arduino-build.yml
@@ -1,5 +1,8 @@
 name: Arduino Firmware Build
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+  
 on:
   pull_request:
     paths:
--- a/.github/workflows/ci-arduino-build.yml
+++ b/.github/workflows/ci-arduino-build.yml
@@ -36,9 +36,11 @@ jobs:
           arduino-cli lib install ACAN2515
           arduino-cli lib install Streaming
 
-      - name: Compile representative examples for Arduino Uno
+      - name: Compile all examples for Arduino Uno
         run: |
-          arduino-cli compile --fqbn arduino:avr:uno --library . ./examples/VLCB_empty
-          arduino-cli compile --fqbn arduino:avr:uno --library . ./examples/VLCB_1in1out
-          arduino-cli compile --fqbn arduino:avr:uno --library . ./examples/VLCB_4in4out
-          arduino-cli compile --fqbn arduino:avr:uno --library . ./examples/VLCB_SerialGC_empty
+          for dir in ./examples/*; do
+            if [ -d "$dir" ] && ls "$dir"/*.ino >/dev/null 2>&1; then
+              echo "Compiling $dir"
+              arduino-cli compile --fqbn arduino:avr:uno --library . "$dir"
+            fi
+          done
--- a/.github/workflows/ci-arduino-build.yml
+++ b/.github/workflows/ci-arduino-build.yml
@@ -1,0 +1,44 @@
+name: Arduino Firmware Build
+
+on:
+  pull_request:
+    paths:
+      - examples/**
+      - src/**
+      - library.properties
+      - .github/workflows/ci-arduino-build.yml
+  push:
+    paths:
+      - examples/**
+      - src/**
+      - library.properties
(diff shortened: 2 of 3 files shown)


SvenRosvall:
Thanks for this Bruno.
Can you split this PR in two? 
One for the serial GC. This is straight-forward. Obviously need some automated tests for the example sketches. That would have exposed those things we missed.
The other one would be for the GitHub build stuff. It would be good if you could provide some links that describe how the Github build stuff works. That would help me understand what your YAML files are doing.

brocci:
Hello Sven. Sure, I could do that split (I hope I will not mess up everything!).
In fact, I discovered the bug in SerialGC during the execution of the workflow.
I think there are pages in Github that describe how to create a workflow. I will try to find the most relevant, or some form of tutorial. It is relatively easy to understand. Basically you describe what to do when something is pushed on the repo, for example, which action/job to execute, what it takes as an input, what it does with the files and what it produces.
I have reused some of the yaml I have been using in other projects or at work.

brocci:
I think the split is OK. I have added a link to the new PR in this one.

brocci:
I have found those pages on Github:

Quickstart: https://docs.github.com/actions/quickstart
Understand workflows: https://docs.github.com/actions/using-workflows
Events and triggers: https://docs.github.com/actions/using-workflows/events-that-trigger-workflows

Using only a subset of the available actions is generally enough to perform the simple tasks we need: build, test, generate doc. The most difficult part if often to find the proper syntax or command line options. 
(I had troubles with CMake for example, because this is a tool I do not use)

Unless explicitely specified, the workflows are executed on a Linux-based virtual machine (runner). So the commands are linux style.

I will try to find some time to create a small document explaining the workflows.

SvenRosvall:
Looked at the report from running your actions in your fork. There is a warning that some of the actions you use are going to be deprecated today. Have you seen this?

https://github.com/brocci/VLCB-Arduino/actions/runs/27510749789/job/81310240980

> Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, arduino/setup-arduino-cli@v2. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

brocci:
Hello Sven,
Yes, I am aware of this. Github is slowing upgrading their runners to NodeJs24 and they are warning users about this change.

It is just a matter of changing the version of the action, for example replacing _actions/checkout@v4_ by _actions/checkout@v5_. No major change.

I could add this _FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true_ line to check the NodeJS20 actions could be executed on NodeJS24 when it will be effectively deployed (but we don't know exactly when).

The alternative is to wait until the workflow report errors, then we will need to update actions in the yaml files.

I will test on my fork what happens when the _FORCE_JAVASCRIPT_ACTIONS_TO_NODE24_ is added, and if it is OK, I will commit the change. I believe this is the safest way.

brocci:
Looks OK on my side.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,49 @@
...
+          sudo apt-get update
+          sudo apt-get install -y doxygen graphviz
+
+      - name: Generate API documentation
+        run: |
+          doxygen Doxygen.Sketch.conf
+          doxygen Doxygen.Library.conf
+
+      - name: Upload sketch docs artifact
+        uses: actions/upload-artifact@v4
```
SvenRosvall, line 40: What does the upload-artifact action do? Where is the artifact uploaded to? Are the updated files committed to the repository?

brocci:
The _upload-artifact_ action copies the generate html documents into the workflow artifacts space:

artifacts

(From the Github top menu -> Actions -> Select Docs workflow -> Select last run -> Look for the Artifacts section)

From there you can download them as ZIP files for review.

The workflow does not commit the files automatically, on purpose. In general, documentation files should be transfered to the project pages. This would require an additional job after the doc generation.

I think it should be possible to fully automate the docs generation and pages updates. So the docs/*.md, html.sketch and html.library files are published on github page in one go, only when something is updated on the main branch. On the other branches, only artifacts would be generated for review.

Let me know if you want to do that.

SvenRosvall:
OK, I see how it works. However I am not sure I agree with the benefits of artifacts that you have to download yourself. The generated files are committed to html.sketch and html.library so they are always available in the library. OK, we have to remember to run Doxygen. Also the API documentation is available on the GitHub "Pages". See https://svenrosvall.github.io/VLCB-Arduino/html.library/hierarchy.html
This is where people go if they want to read the documentation.

I am happy with the other two workflows. I don't have a problem including this doc workflow for now. We can look at how to work with Doxygen later. Happy to approve this pull request.

SvenRosvall (approved):

brocci:
Thank you. Workflows could be improved over time and new ones created if need arise. 

I believe my docs.yaml workflow should be improved as discussed. The objective being that you would no longer need to run Doxygen manually and commit the generated documents. The pages would be updated automatically during the execution of the workflow. But this must only happen on the main branch. Any update on other branches should not trigger this as we could be making several changes in a row and until they are merged into main, we don't want the "official" documentation on github pages to be updated. Consider the artifacts as "preview".

SvenRosvall:
Bruno, I don't have your email address so sending this via GitHub. I hope it works even if this is a closed PR. My email is sven@rosvall.ie

The last update of the docs workflow where you fixed the environment didn't
get run when I merged. I started it manually and it ran fine. Good.

It did show the message about NodeJS 20 being deprecated. It complains
about the action checkout@v4. I am curious why you fixed this in other
workflows with an environment variable instead of using a later version of
the checkout action? There seems to be a few updated versions of checkout.
Note that checkout@v5 fixes the NodeJS version problem.
See https://github.com/actions/checkout/blob/main/README.md

For the action upload-artifact the readme file doesn't describe the
versions. But it uses upload-artifact@v7 in the examples. So the @v4 that
is used in our workflowas seems to be old.
See: https://github.com/actions/upload-artifact
Click on "Releases" to see other versions and brief descriptions. 

Cheers / Sven

Issue #215, opened 2026-06-13
Start-of-Day in VLCB_4in4out example sketch


SvenRosvall:
The Start-of-Day concept needs an example. Add it to the 4in4out sketch as it has a few inputs to report status for.

Once TimedResponse feature is implemented, convert this SoD implementation illustrate how to do SoD using TimedResponse.

brocci:
I had a look in the forum and VLCB documents, but I haven't been able to find information about the "Start of Day" concept.
Could you please provide some pointers?

SvenRosvall:
Yes, SoD is a strange concept. It is not a core part of CBUS (or VLCB). Instead it is a concept implemented with what is available in CBUS.

The best description I know of is in https://merg.org.uk/merg_wiki/doku.php?id=cbuscentral:sod

brocci:
Thanks, Sven. I wouldn’t say “strange,” maybe more “interesting.” I understand the need to start from a known state on a big layout. It might be a workaround to fill a gap, but ultimately, if it works, it works.

SvenRosvall:
Feature implemented.

PR #218, opened 2026-06-15
Implement Start-of-Day event in 4in4out example

SvenRosvall:
This is a naive implementation. See #215.  I want to get this working before I implement it with TimedResponse. Otherwise if something fails I don't know if it is the SoD code or the TimedResponse code that failed.

There is a question in the code if we should send SoD response events for inputs that can only be one of ON or OFF. Current code does, which means those events will always be ON or always OFF, which I doubt the usefulness of.

### DIFF
--- a/docs/SketchDevelopment.md
+++ b/docs/SketchDevelopment.md
@@ -74,6 +74,21 @@ The slot event strategy reqires a larger event table but fewer event variables.
 Total memory consumption for the events table is very similar to classic
 events strategy.
 
+### Start-of-Day Event
+A Start-of-Day event is optional an event that is consumed by the node that triggers
+sending out a number of events to indicate the state of inputs on the node.
+These status events are used to update consumers on the layout that depend
+on the state of these inputs.
+An example for this is a control panel that needs to light up the correct
+indicator lights when powered up.
+
+A Start-of-Day event is an ordinary event that a node may consume and trigger
+the sending of status events.
+It is usually recognised with a special value in one of the event variables.
+
+The example sketch VLCB_4in4out.ino illustrates how a Start-of-Day event
+can be implemented. See its [README](VLCB4in4out_README.md#event-variables) file
+
 ## Choosing a module identifier
 A module identifier consists of a _Manufacturer ID_ and a _Module ID_.
 The combination of these two must be unique.
--- a/docs/VLCB4in4out_README.md
+++ b/docs/VLCB4in4out_README.md
@@ -94,7 +94,10 @@ OP_CODE | HEX | Function
 ### Event Variables
 
 Event Variable 1 (EV1) identifies a consumed only event if its value is zero.
-A non-zero value will identify the source (in this case switch) for a produced
+A value of 255 identifies the event as a request for Start-of-Day reporting,
+i.e. send an event for each input to update control panels etc, of the current
+state.
+A value of 1 to 4 will identify the source (in this case switch) for a produced
 event.
 
 ### Consumed Events
--- a/docs/VLCB4in4out_README.md
+++ b/docs/VLCB4in4out_README.md
@@ -94,7 +94,7 @@ OP_CODE | HEX | Function
 ### Event Variables
 
 Event Variable 1 (EV1) identifies a consumed only event if its value is zero.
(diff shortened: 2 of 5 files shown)


MartinDaCosta53 (commented):
Line 366 in VLCB_4in4out.ini should read 
for (byte i = 0; i < NUM_SWITCHES; i++)

Regarding the TODO at line 378, I don't see how you can do a meaningful SoD for anything that is not a bi-stable input like a toggle switch.  We don't store the EoD state of a toggle button and so don't know how it was left at the last EoD.  Similarly, ON only and OFF only are meaningless from a SoD point of view.

Do we want to send the actual event at SoD or just report it's state with a query response opcode?  I think the former, which is what you have done.

SvenRosvall:
Good point about toggle switches. Duncan made a comment recently about saving user data in eprom. There isn't much help for the user today for this. I'll add a feature request for user eprom data and that the toggle switch is a good use case for this.

Yes the intention was to show an example for SoD with a sequence of status events.
Agree that the event request mechanism is a better way than SoD. I have used it successfully with JMRI. But I have no idea on how to do that with CBUS modules. I guess this is why people use SoD instead.

And thanks for your thought on ON/OFF only switches. I'll update the code.

dpharris:
Maybe I am misunderstanding, but i understand SoD to report the current
state on a module. Therefore, it should teoort saved state or the result of
initial reads of inputs. This would include On-only,since other nodes have
no knowledge that these are On-only, so need to see that result.

A node could always have one or both a SoD-events and a SoD-reports
capability, by offering those capabilities in the app code.

I think that there should definitely be a capability to store user-data,
beyond NV, events, and EVs.

David

SvenRosvall:
Updated the code with Martin's comments. 
Added issue #219 for remembering state of toggle switch.
For now, the toggle switch is included. If it hasn't been operated at the time of a SoD event, the default state will be reported.

SvenRosvall:
David, the ON-only switch and OFF-only switch doesn't make sense to send status events for. They will always report ON or OFF respectively. So what can a consume make of this? 
As I understand these switches they will always operate in pairs. One button sets a turnout to normal and the other button sets it to reversed. The turnout will be set the way the last of these two buttons were pressed. If we include these buttons in the SoD events then you get an event for either way for the turnout. The timing of arrival of these events will determine which way the turnout will be set. Totally random. 
Note also that these two buttons can be connected to different boards so it is impossible for a node to maintain knowledge of which button was pressed last. Also the node doesn't know which two buttons operate in pair, even if they are connected to the same board.

dpharris:
Ah, I see the problem.  When using buttons, the consumer node would
need to maintain the state, not the one or two producer nodes.  That
seems a little messy.  Perhaps producer nodes should keep its state by
also monitoring the opposite event, ie an ON-only producer should
change its state to Off if an Off-event is received?
This leads to a general discussion of how to use ON/OFF, ON-only, and
Off-only modes.

MartinDaCosta53:
In the case of an ON only and OFF only button controlling a point, the only sensible SoD would be to interrogate the feedback event from the point.  This would then set the tell-back light on the control panel.  Of course, I have made assumptions here about how the layout works, but no other SoD mechanism makes sense, IMO.

With regard to an ON/OFF toggle button controlling the point, there is a need to sync the producer status with the actual point position at SoD.  That can only be done by operating the button, I would suggest.

SvenRosvall:
The intention here is not to create a perfect SoD implementation, but to illustrate how SoD could be implememented.
As both of you have highlighted this is tricky to design correctly in all cases. Even if we implement these concerns there will be many more edge cases. Chasing these will just take too much effort.

MartinDaCosta53:
I agree.  Most of this discussion needs to be in the Designer's Guide.  Our examples should only show an implementation for that which is relevant.  SoD implementation will be unique to every layout.

dpharris:
Yes, the library should enable the user to implement a SoD, but not specify
the details.
For example, using Slots the user would probably designate a slot to be the
SoD slot. If it is consumed, then the node would be programmed to take some
action, like report state.

David

MartinDaCosta53 (approved):

PR #220, opened 2026-06-17
Serial Stream Refactor and Split-Channel Example

### LINKED ISSUE #216: Using both SerialGC and SerialUserInterface
I came across this note in Examples.md:

`Note that the serial examples do not use the SerialUserInterface as it would interfere with VLCB message communication.`

This means that you cannot have a VLCB node sketch that uses both **SerialGC** and **SerialUserInterface** because they use the same _Serial_ type, which is HardwareSerial on an Uno for example.

One way to get around this would be to replace _Serial_ with _Stream_ in the constructors. This would provide more flexibility, for example:
- SerialGC could use Serial(HardwareSerial)
- SerialUserInterface could use SoftwareSerial(Bitbanging serial)
- Both will be able to coexist

This would be backward compatible and would also work on boards with multiple UARTs, like an Arduino Mega.

Note: I have some code ready, but I need to perform tests.

brocci:
This PR refactors serial bindings to use Arduino `Stream` instead of hard-wiring to global `Serial`, while preserving default behavior.

It also adds a new example showing split serial channels:
- `SerialGC` on hardware `Serial` (GridConnect transport)
- `SerialUserInterface` on `SoftwareSerial` or Serial1 on Mega2560 (user console)

## Changes
- Refactor `SerialGC` constructor/member type to `Stream&`
- Keep transport stream protocol-clean in `SerialGC::begin()` (Removed startup banner print from begin)
- Refactor `SerialUserInterface` to injected `Stream&` with default `Serial`
- Add new split-channel example: `examples/VLCB_SerialGC_SerialUI_SoftwareSerial/VLCB_SerialGC_SerialUI_SoftwareSerial.ino`
- Document new example in `docs/Examples.md`

## Why
- Allow `SerialGC` and `SerialUserInterface` to work with either `HardwareSerial` or `SoftwareSerial` (or other Stream-compatible ports).
- Enable practical split-channel use where protocol traffic and UI commands are isolated.
- Avoid accidental non-GridConnect text on the transport stream.

## Compatibility
- Existing sketches remain compatible via default constructors using `Serial`.
- Behavior is unchanged unless an alternate stream is explicitly passed.

Closes #216


SvenRosvall:
Quick early comment (before reviewing thoroughly) is that the name of the sketch is long. Also I'd like to include "empty" in the name to show what functionality it provides. 

Suggest "VLCB_SerialGC_empty_withSerialUI.ino"
OK, that is not much shorter. :( The "SerialGC" must be there. Or can we shorten this marker? (applies to all serial example sketches) "VLCB" must be there to avoid confusion with the examples in the "CBUS" library if you work with both libraries. 

Does this make sense?

brocci:
Yes, it makes sense. The longest example filename is `VLCB_long_message_example`, 25 characters.
Would `VLCB_SerialGC_SerialUI_empty` be better?

SvenRosvall:
> Would `VLCB_SerialGC_SerialUI_empty` be better?

Yes. it does. Just thinking that the "empty" bit should be earlier in the file name. But that would affect existing VLCB_SerialGC sketches. Need to think about a naming convention and if needed change all sketches.

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
```diff
@@ -0,0 +1,144 @@
+//  Copyright (C) Sven Rosvall (sven@rosvall.ie)
```
SvenRosvall: Please claim copyright for yourself.

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
```diff
@@ -0,0 +1,144 @@
...
+/// print code version config details and copyright notice
+//
+void printConfig()
+{
+  // code version
+  serialUserPort << F("> code version = ") << VER_MAJ << VER_MIN << F(" beta ") << VER_BETA << endl;
+  serialUserPort << F("> compiled on ") << __DATE__ << F(" at ") << __TIME__ << F(", compiler ver = ") << __cplusplus << endl;
+
+  // copyright
+  serialUserPort << F("> © Sven Rosvall (MERG 3777) 2026") << endl;
```
SvenRosvall: Claim copyright for yourself. ;)

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
```diff
@@ -0,0 +1,144 @@
...
+  serialUserPort << F(", CANID = ") << VLCB::getCANID();
+  serialUserPort << F(", NN = ") << VLCB::getNodeNum() << endl;
+
+  printConfig();
+}
+
+//
+/// print code version config details and copyright notice
+//
+void printConfig()
```
SvenRosvall: Can you put this function at the bottom of this file. Then it is easier to compare the different "*_empty" sketches.

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
brocci: Ahah. I didn't think about this ;-)

test/ArduinoMock.cpp
```diff
@@ -99,64 +99,8 @@ byte lowByte(unsigned int i)
...
-
-void Serial_T::println(const char *)
-{
-}
-
-void Serial_T::flush()
-{
-}
-
-Serial_T & operator<<(Serial_T & s, int i)
```
SvenRosvall, line 128: I implemented these operators this way so that any debug info would be printed in the text outputs for the unit tests. This is handy when troubleshooting.

Can you keep these operators but change to use your Stream class?

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
brocci: Well, 90% of the code is yours. But OK ;-)

examples/VLCB_SerialGC_SerialUI_empty/VLCB_SerialGC_SerialUI_empty.ino
brocci: I thought I had put it back in my last commit. Need to check.

Arduino/Arduino.h
```diff
@@ -28,15 +29,4 @@ void pinMode(int, PinMode);
 void digitalWrite(int, PinState);
 byte digitalRead(int);
 
-struct Serial_T
-{
-  void begin(int baudrate);
```
SvenRosvall, line 33: This begin() function is used by the example sketches.

No, these headers are not used by the Arduino IDE or your new workflows when compiling the example sketches. However I also use the headers in the Arduino directory for my IDE (CLion) so that the example sketch codes come up nicely in the IDE without any red markers.  So please keep the begin() function in the Stream.h .

Arduino/Stream.h
```diff
@@ -0,0 +1,24 @@
...
+  virtual ~Stream() = default;
+
+  virtual bool available() { return false; }
+  virtual char read() { return 0; }
+  virtual void print(const char *) {}
+  virtual void println(const char *) {}
+  virtual void flush() {}
+};
+
+// Operator overloads for Stream to work with Streaming library
```
SvenRosvall: These operators don't need definitions in the Arduino directory. This directory is only used to make my IDE work well with Arduino code. 
For executing test code the definitions should reside in ArduinoMock.cpp in the test directory. See also comments there.

test/ArduinoMock.cpp
brocci: I will have a look.

src/SerialGC.h
```diff
@@ -24,7 +25,7 @@ namespace VLCB
   class SerialGC : public CanTransport
   {
   public:
-    SerialGC(typeof(Serial)& _serial = Serial) : serial(_serial) {}
+    SerialGC(Stream& _serial = Serial) : serial(_serial) {}
```
SvenRosvall, line 28: Yes! That looks much nicer. Should have done that from the beginning. :)

src/SerialUserInterface.h
```diff
@@ -20,6 +23,8 @@ class SerialUserInterface : public Service
 {
 public:
   /// @cond LIBRARY
+  SerialUserInterface(Stream& _serial = Serial) : serial(_serial) {}
```
SvenRosvall, line 25: The constructor should be declared above the @cond line.
The @cond tells Doxygen to ignore a block (until @endcond) of declarations. This is useful to reduce the amount of documentation a sketch developer needs to see. (The library developer needs to see everything.) 
This constructor is something a sketch developer will use and thus it should be include in the sketch developer documentation.

SvenRosvall:
Thanks for this pull request. It looks good. Will be a good feature. I know John Fletcher will be happy about this. 

The comments I have added are mostly about my style and how I work with this project. Sorry that this is not documented properly. So the comments are there to share this until I get that bit done.

brocci:
> Thanks for this pull request. It looks good. Will be a good feature. I know John Fletcher will be happy about this.

Thank you. I will fix the remaining "glitches" and commit on this PR. Happy to make John happy :-)

> The comments I have added are mostly about my style and how I work with this project. Sorry that this is not documented properly. So the comments are there to share this until I get that bit done.

No problem.

dpharris (commented):
This looks fine to me.  
In the example, perhaps it would be good show alternatives:
VLCB::SerialGC serialGC(Serial);  // CAN transport object using hardware Serial
//VLCB::SerialGC serialGC(SofwareSerial); // CAN ransport via SoftwareSerial
//VLCB::SerialGC serialGC(Serial1); // CAN transport via Serial1, eg on the Mega, R4, ESP32-S3, Pico

brocci:
> In the example, perhaps it would be good show alternatives: 
VLCB::SerialGC serialGC(Serial); // CAN transport object using hardware Serial //VLCB::SerialGC serialGC(SofwareSerial); // CAN ransport via SoftwareSerial //VLCB::SerialGC serialGC(Serial1); // CAN transport via Serial1, eg on the Mega, R4, ESP32-S3, Pico

Well, there is some information about this in the header. I could make it more explicit.
I am not sure that SoftwareSerial should be used for SerialGC, which I think uses 57600 bauds by default. It becomes unreliable above 19200 bauds on an AVR. It might be OK on other targets like ESP32 or RP2040. But I haven't tested.

SvenRosvall (approved):

brocci:
Would that revised version of the header text be acceptable?

```cpp
// This example demonstrates split serial usage:
//
// - SerialGC uses the default hardware Serial port for GridConnect transport
// - SerialUserInterface uses SoftwareSerial for user commands and status output
// - Optionally, hardware Serial1 can be used for the SerialUserInterface on Mega2560 by setting the USE_MEGA_SERIAL1 build flag in platformio.ini
//
// This is useful on boards where the USB serial port is already committed to the VLCB/GridConnect transport, but a separate command console is still wanted.
//
// Any Stream-compatible serial port can be used for the SerialUserInterface, but SoftwareSerial is used in this example for compatibility with a wide range of boards. 
// Note that SoftwareSerial has some limitations, such as not being able to receive data while transmitting, and may not be suitable for high-speed communication. 
// It is not recommend to use SoftwareSerial for the SerialGC transport, as it may not be able to keep up with the data rate and could lead to data loss.
// For better performance, consider using a board with multiple hardware serial ports and connecting the SerialUserInterface to one of those.
```

If you agree, I will push the commit.

dpharris:
I agree in general re opening up Serial.  
As to the header text, I would perhaps add an introductory sentence or two, perhaps:  
"In order to have concurrent use of the SerialGC and SerialUserInterface classes, the library allows a user to direct these to alternate interfaces in their declarations.  This example demonstrates this split usage."

brocci:
@SvenRosvall , @dpharris
I have updated the header to clarify usage. Hope this is fine now.

dpharris:
Looks great to me.  Do I need to merge, or will Sven do that?

PR #223, opened 2026-06-23
Add missing environment field for deploy-pages

brocci:
In the workflow, `action/deploy-pages@v4` requires the job to declare environment: github-pages, which was missing from the deploy-gh-pages job. The GitHub API rejects the deployment creation without it.

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -53,6 +53,8 @@ jobs:
     needs: [generate-doxygen]
     if: github.event_name != 'pull_request' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
     permissions:
       pages: write
       id-token: write


brocci:
@SvenRosvall I have created this new PR with the fix for the GH Pages. But I cannot test it until it is merged to `main`

brocci:
@SvenRosvall If you are OK with the docs workflow, I think you could remove the html.sketch and html.library folders from the repo. They won't be updated, unless you explicitely generate a new version locally and commit them, which is no longer required.

SvenRosvall:
This is a good idea. To be on the safe side I wanted to leave it for a while.

SvenRosvall:
Also want to let you know that your workflows did highlight a coding error I did in a recent change. It was in an example sketch which I normally don't check in the Arduino IDE. Many thanks for this work. It is already showing its value.

brocci:
Glad to see it brings some value. Thank you.

PR #227, opened 2026-07-01
Updates for NodeJS 24.

SvenRosvall:
Updated some workflow actions to versions that work with NodeJS 24.

### DIFF
--- a/.github/workflows/ci-arduino-build.yml
+++ b/.github/workflows/ci-arduino-build.yml
@@ -1,6 +1,7 @@
 name: Arduino Firmware Build
 
 env:
+  # Fix for NodeJS 24 until action arduino/setup-arduino-cli is fixed.
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
   
 on:
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -1,8 +1,5 @@
 name: Docs
 
-env:
-  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
-
 on:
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-arduino-build.yml
brocci, line 1: My understanding is that currently the github runners still have node20 enabled. The env variable just forces the action to execute on node24. When the env variable is not present, the action will execute on the node version specified in its code. The current arduino-cli action is tight to node20 and there is a PR pending to move it to node24. If the env variable is removed and arduino-cli is not updated on time, when node20 will be disabled on runners, then the workflow will fail. So yes, better to leave it for now.

.github/workflows/ci-docs.yaml
brocci, line 1: If you have tested it working and pages are updated OK, then that's fine. Make sure you have enabled the Pages option to use the workflow. Then remove the html folders from the repo.

.github/workflows/ci-host-tests.yml
brocci, line 1: If it works, it works

.github/workflows/ci-docs.yaml
SvenRosvall: Yes, I have tested them. Even tested deploying the pages by temporarily removing the condition for branch name.
Will remove the html folders.

brocci: Good. The side effect is that the project would appear as C++ instead of HTML ;-)

SvenRosvall: Ah, didn't think of that. ;)

brocci (approved):

PR #232, opened 2026-08-28
Add functions to send any VLCB message.

SvenRosvall:
As requested by Nick Locke. He needs to send AREQ messages. These messages shall not include the sending node number.


SvenRosvall:
Replaced by PR #233.

